PR #3, opened 2015-05-08
Add shortanswer directive

bnmnetp:
Adapting the journal directive from acbart to work in the current runestone environment.


bnmnetp:
Needs ajax on the server to retrieve answers, and instructor interface for grading and providing feedback.

PR #7, opened 2015-06-09
Changes to improve the timed exam feature

ericsonga:
I changed it from requiring 4 directives to 3 and added the ability to
optionally turn off feedback and turn off the results.


ericsonga:
The audio tours aren't being logged correctly.  This should fix it.  The audio tours also don't play sometimes and this will tell people to close the window and try again.

PR #8, opened 2015-06-29
Remove the duplicate bookfuncs.js script tag in built page's head

mmoldavan:
the bookfuncs.js script was getting added to build page's html <head> twice, and that was causing a lot of javascript to be executed twice unnecessarily.

It was being added in two places: activecode/activecode.py's setup() and datafile/__init__.py's setup(). Removed it from the later.

### DIFF
--- a/runestone/datafile/__init__.py
+++ b/runestone/datafile/__init__.py
@@ -23,7 +23,6 @@ from docutils.parsers.rst import Directive
 
 def setup(app):
     app.add_directive('datafile',DataFile)
-    app.add_javascript('bookfuncs.js')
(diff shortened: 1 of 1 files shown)


bnmnetp:
Thanks for the PR.  I have to think about this a little harder before accepting.

In the old days (last year) this didn't matter, and was in fact beneficial because now if a page only has a datafile, but no activecode then bookfuncs will not be loaded at all!

A better solution would be to have add_javascript detect a duplicate add and not do the add.

Or, move bookfuncs.js to the (layout.html) template so it is *always* included and remove it from both.

Brad

PR #9, opened 2015-07-02
Ported the assess components and the datafile component to be RSE 0001 complient

riknos314:
Created new js files for each and updated the python for Sphinx to output the new intermediate html.  There is a test_overview.rst file in /common/project_template/_sources for testing purposes.  
Each of the assess and datafile directories has a readme file with example syntax of the intermediate html and the associated options.


bnmnetp:
This is great work.  I would like to stay backward compatible with the existing sphinx directives.

I think if you make mchoicemf, mchoicema, and random subclasses of the current mchoice it should be super easy to keep these directives around.  But, you should have them spit out a message that they are Deprecated and should be changed to mchioice with the appropriate option.  This way we don't break all of the existing writing.

riknos314:
The directives have been updated to add backward compatibility with the existing format for MC questions.

Timed and Fill in the blank are not backward compatible and I'm not sure that there is a simple way to make them so.  

The directive for Datafile has not changed at all in restructured  text, and thus is backward compatible.

PR #10, opened 2015-07-06
Fixes to audio tour logging and suggestion to close the window if it doesn't play

ericsonga:
The audio tours weren't logging correctly.  I have fixed that.  I also changed the string that says that the audio tour is loading to recommend closing the window if they don't play and trying again.


bnmnetp:
I Incorporated your changes to audio tours in the new implementation.  I think the others are already done by the guys.

PR #34, opened 2015-07-10
Ported reveal to be RSE-0001 complient

riknos314:
As per #24.


bnmnetp:
The three active PR's all contain implementations of the RunestoneBase and the logging functions.  These should come from a common file, and I need to restore the logging back to the server functionality.

riknos314:
As is, the logging functions defined in RunestoneBase aren't even being called (in fact, nothing related to RunestoneBase is used at all).  We just mirrored the format that you had set in your javascript implementation of activecode.  I'd be more than happy to work on a common RunestoneBase file, if you would let me know what functionality RunestoneBase should provide.

riknos314:
I removed the duplicate declarations of runestonebase and created a common runestonebase file (you'll find that in the PR for tabbedstuff).

PR #38, opened 2015-07-14
re-added the textfield directive

riknos314:
The textfield directive was previously included as part of the assess components, but is no longer used there.  Since it can be used by activecode, we have moved it into the activecode directory.


bnmnetp:
Can you limit this PR to only have the new textfield.py ?   All of the changes you made in activecode.py are going to conflict with my version.

riknos314:
I just changed activecode.py back to the state that it was at in the previous commit (3a6ebadac64d3b4f45b32dd7f364ee5bba47d52e).

Issue #49, opened 2015-07-20
Strange radio button placement on some multiple choice


bnmnetp:
The sentence should start on the same line as the radio button.

![screenshot 2015-07-20 14 13 31](https://cloud.githubusercontent.com/assets/51115/8784876/aa9cb8c4-2ee9-11e5-99b9-27dce75819cb.png)

riknos314:
Could you please provide the raw html of a question that displays this behavior?

bnmnetp:
Here it is:
```html
            <ul data-component="multiplechoice" data-multipleanswers="false"  id="question1_2_2">intro-3-2:  What is the difference between a high-level programming language and a low-level programming language?
            
            <li data-component="answer"  id="question1_2_2_opt_a">It is high-level if you are standing and low-level if you are sitting.</li><li data-component="feedback" id="question1_2_2_opt_a">In this case high and low have nothing to do with altitude.</li>
            
            <li data-component="answer"  id="question1_2_2_opt_b">It is high-level if you are programming for a computer and low-level if you are programming for a phone or mobile device.</li><li data-component="feedback" id="question1_2_2_opt_b">High and low have nothing to do with the type of device you are programming for.  Instead, look at what it takes to run the program written in the language.</li>
            
            <li data-component="answer" data-correct id="question1_2_2_opt_c">It is high-level if the program must be processed before it can run, and low-level if the computer can execute it without additional processing.</li><li data-component="feedback" id="question1_2_2_opt_c">Python is a high level language but must be interpreted into machine code (binary) before it can be executed.</li>
            
            <li data-component="answer"  id="question1_2_2_opt_d">It is high-level if it easy to program in and is very short; it is low-level if it is really hard to program in and the programs are really long.</li><li data-component="feedback" id="question1_2_2_opt_d">While it is true that it is generally easier to program in a high-level language and programs written in a high-level language are usually shorter, this is not always the case.</li>
            

            </ul>
```

But, oops, it is like this in the current production system as well.  So I guess this isn't very high priority.

riknos314:
That looks good, but I really wanted the HTML of the elements produced by the javascript, not the HTML going into the javascript.  I want to see what is causing the output to have that spacing because I've never encountered that before.

bnmnetp:
If you know of some way to get the source that has been generated by javascript, let me know and I will do it.  Otherwise I'm not sure.  I gave you what I got from doing a show source in Chrome.

[...]

PR #52, opened 2015-07-20
updated jsComponents to use this.logBookEvents

### LINKED ISSUE #48: all components should use this.logBookEvent defined in RunestoneBase

riknos314:
Resolves #48


riknos314:
I forgot to make sure that runestoneBase was being inherited in the previous commit.  This has now been fixed.

Also includes a solution for #49 that fixes the weird line-wrapping that would happen if the answer text of an MC question was long enough to be wider than the container that it is contained in.

PR #86, opened 2015-08-07
Fixes to timed exams.

ericsonga:
Changed the start button to green.  Changed the arrows for next and
previous to single arrows.  Moved the next and previous arrows to below
the question.  Moved the question number jump buttons to below the next
and previous buttons.  Updated the results to include the question
numbers.


bnmnetp:
Did you make changes to bootstrap.min.css?  The system seems to think you did but is not showing me what was changed.

PR #97, opened 2015-10-03
Save code during sphinx build for use in grading

presnick:


presnick:
Now also includes new usageassignment directive, which is the start of being able to author assignments as .rsts and have them show up as gradeable assignments on the website. So far, includes only "use" assignment type, where it monitors how much time and how many of the activities the student engaged with.

runestone/activecode/activecode.py
```diff
@@ -126,7 +127,8 @@ class ActiveCode(Directive):
         'timelimit': directives.unchanged,
         'stdin' : directives.unchanged,
         'datafile' : directives.unchanged,
-        'sourcefile' : directives.unchanged
+        'sourcefile' : directives.unchanged,
+        'available_files' : directives.unchanged
```
bnmnetp: Is this a list of div_ids for datafiles that have been saved or what should the value of this be?

presnick: Yes. For example, see https://raw.githubusercontent.com/presnick/ProgramsInformationPeople/master/_sources/Assignments/ps3.rst  ps_3_7

runestone/usageAssignment/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+
+from docutils import nodes
+from docutils.parsers.rst import directives
+from docutils.parsers.rst import Directive
+from sqlalchemy import create_engine, Table, MetaData, select, delete
+from sqlalchemy.orm import sessionmaker
+
+
+def setup(app):
+    app.add_directive('usageassignment',usageAssignment)
```
presnick: For usage examples see directives near the top of https://raw.githubusercontent.com/presnick/ProgramsInformationPeople/f15/_sources/Assignments/ps5.rst

runestone/activecode/activecode.py
```diff
@@ -229,6 +232,60 @@ def run(self):
...
+            else:
+                source = "\n".join(self.content)
+                suffix = "\n"
+        else:
+            source = '\n'
+            suffix = '\n'
+        try:
+            engine = create_engine(env.config['dburl'])
+            meta = MetaData()
+            course_name = env.config['course_id']
```
bnmnetp: course_id should already be available from the template arguments in pavement.py, we should not duplicate this bit of information in conf.py.

presnick: I see the project_name variable in pavement.py, which is stored in the build option object. Can you tell me how to access that build option object inside an extension like activecode.py?

bnmnetp: ``env.config.html_context['course_id']``   Everything from template_args in the pavement.py file gets into there.

However, I'm trying to do a build and even though I added a line to my conf.py with a value for dburl I'm getting an exception that 

```
The exception is  No such config value: dburl
Unable to save to source_code table in activecode.py. Possible problems:
  1. dburl or course_id are not set in conf.py for your book
  2. unable to connect to the database using dburl

This should only affect the grading interface. Everything else should be fine.
```

I added a line to print out the exception.  How did you add the dburl to your conf.py file?

bnmnetp: Note, I think I would rather add the dburl to pavement.py than conf.py as I currently do not need to customize conf.py for each new course that is built by the course building interface.

bnmnetp: Yep, I double checked my comment, and the conf.py file is not saved/customized for each custom built course.  So, adding both the dburl to pavement.py and using the course_id that is already in pavement.py is a much better way to go from my perspective.

presnick: OK. Will switch to using pavement.py for this.

PR #108, opened 2015-11-26
Localization and Finnish translation

aleksi-lukkarinen:
Hello,

I'm working in a project, in which we are producing instructional material in Finnish and thus need our platform in Finnish as well. We are going to translate at least several of Runestone's components, but this pull request contains translation only for ActiveCode, as well as some general plumbing for passing parameters from Paver's config file etc. We would like to have these changes to be included into Runestone itself, so please review them and tell, if this approach is sufficient or if you want something to be changed. Certainly the localization could be done technically better, and there are JavaScript libraries for it, but I thought you probably wouldn't want any (or more) external dependencies, so in addition to abandon the idea of utilizing them, I decided to keep things as simple as possible. Translations for new languages are easy to add, if the current simplistic implementation is good enough for them.


bnmnetp:
Thanks for this, I will take a closer look at it over the next few days.

Brad

runestone/__main__.py
```diff
@@ -30,6 +30,8 @@ def init():
     conf_dict['author'] = click.prompt("Your Name ", default=getpass.getuser())
     conf_dict['project_title'] = click.prompt("Title for this project ", default="Runestone Default")
     conf_dict['python3'] = click.prompt("Use Simple Python3 Semantics ", default="false")
+    conf_dict['language'] = click.prompt("Locale ", default="en")
+    conf_dict['hide_control_captions'] = click.prompt("Hide control captions? ", type=bool, default=False)
```
bnmnetp: I can't tell what the hide_control_captions option is supposed to be for.  Can you tell me a little about its purpose?

aleksi-lukkarinen: Hi and thanks for reply!

The hide_control_captions isn't directly related to the translation, but it is meant to be a global flag to hide captions of all controls in case they are unnecessary obtrusive, like they seem to be in our case. I mean, if a caption is not refered from the body text and doesn't provide any useful information to the reader, it's useless. It could be useful to create an instance-based flag for each control, too, but this is to be a global flag so that there wouldn't be the need to add instance-based flags everywhere throughout the produced material in case all the captions are to be hidden. Of course, only the ActiveCode control reacts to it for now via a simple if clause. I'm sorry I didn't put it into another pull request.

Aleksi

runestone/activecode/js/activecode.js
```diff
@@ -246,28 +627,31 @@ ActiveCode.prototype.createOutput = function () {
...
 
 ActiveCode.prototype.addCaption = function() {
     //someElement.parentNode.insertBefore(newElement, someElement.nextSibling);
-    var capDiv = document.createElement('p');
-    $(capDiv).html(this.caption + " (" + this.divid + ")");
-    $(capDiv).addClass("ac_caption");
-    $(capDiv).addClass("ac_caption_text");
 
-    this.outerDiv.parentNode.insertBefore(capDiv, this.outerDiv.nextSibling);
+    if (eBookConfig.hideControlCaptions == false) {
```
aleksi-lukkarinen: Here is the place where the hide_control_captions flag is currently effective.

bnmnetp:
Accepted another PR that uses the query $i18n library.  We should go in that direction moving forward.

PR #114, opened 2016-01-16
Fix to read audio files from the server

ericsonga:
Reading the audio files from the server rather than from the
../static/audio directory

### DIFF
--- a/runestone/activecode/js/activecode.js
+++ b/runestone/activecode/js/activecode.js
@@ -987,7 +987,7 @@ AudioTour.prototype.tour = function (divid, audio_type, bcount) {
         // akey+".mp3' type='audio/mpeg'><source src='http://ice-web.cc.gatech.edu/ce21/audio/"+akey+
         // ".ogg' type='audio/ogg'>Your browser does not support the audio tag</audio>";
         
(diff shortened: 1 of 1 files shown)


runestone/activecode/js/activecode.js
```diff
@@ -986,11 +986,14 @@ AudioTour.prototype.tour = function (divid, audio_type, bcount) {
         // str+="<audio id="+akey+" preload='auto'><source src='http://ice-web.cc.gatech.edu/ce21/audio/"+
         // akey+".mp3' type='audio/mpeg'><source src='http://ice-web.cc.gatech.edu/ce21/audio/"+akey+
         // ".ogg' type='audio/ogg'>Your browser does not support the audio tag</audio>";
+        
+        var dir = "http://media.interactivepython.org/studentcsp/audio/"
```
bnmnetp: Barb,

We want to use ``eBookConfig.basecourse`` here instead of hardcoding the path to include studentcsp

ericsonga: Okay I have changed it to use that.  I had to add base course to pavement.py for each book as well.

PR #116, opened 2016-02-03
Master

ericsonga:
This is to fix the problem of multiple logs of timed exam answers.  The first time the user took a timed exam the answers are logged correctly, but if the user goes back to look at the answers again they were being logged again and incorrectly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ nosetests.xml
 .idea
 bower_componenets/
 test/
+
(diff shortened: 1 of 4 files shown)


ericsonga:
This started when your students changed the timed code.  They didn't handle the case when the answers are displayed again when you go back to view the timed exam again after you finished it

ericsonga:
This isn't a work in progress.  I have tested it.

PR #117, opened 2016-02-29
Changes to clickable area to show table border and changes to allow for multiple distractors in Parsons problems

ericsonga:
I modified the clickable area css to show the table borders and made changes to allow an option on Parsons problems to allow the maximum number of distractors to be specified.


bnmnetp:
Barb,

I think this might be missing something.  I cannot find anything related to Parsons code in this PR.

Also it says that there are no changed files, but there are two commits.  The commits are old and are confusing me.  I think we might need to talk through this to make sure that this PR isn't accidentally undoing some good stuff that has been done.

ericsonga:
This is odd.  I had made changes.  Let me see what happened.

ericsonga:
I did a pull request from the wrong version.  This was from my older version and not the right one.  Ignore this one and take the new pull request.

PR #120, opened 2016-03-09
call gethist.json instead of gethist so that we get the right headers…

presnick:
… back


bnmnetp:
Paul, this conflicts with my change to the same line to use getJSON.  which also fixes the problem for me in Chrome.  But, I'll do both since its such an easy fix and avoid dealing with this conflict and clearly adding the .json suffix is a good thing.

PR #122, opened 2016-04-03
Added server storage functionality to assess questions

flysaiah:
Multiple choice, FITB, drag-n-drop, clickable area, and timed exams now all check the server first to load from, then if the server doesn't return anything, they proceed to load from local storage as they always have.


bnmnetp:
One thing that I don't see, but that we need to add takes into consideration the following scenario.

1.  The student has been working on their laptop, and has a long history of answers in their local storage.
2.  The student goes to a new computer, and the local storage is populated from the server.
3.  The student then updates or changes their answer on the new computer which updates the server as well as the local storage on the new computer.
4.  The student goes back to the old computer, now local storage should be updated here as well, as long as the timestamp on the items from the server are newer.

runestone/assess/js/mchoice.js
```diff
@@ -284,64 +273,109 @@ MultipleChoice.prototype.randomizeAnswers = function () {
...
-    if (len > 0) {
+};
 
-        var ex = localStorage.getItem(eBookConfig.email + ":" + this.divid);
-        if (ex !== null) {
-            var arr = ex.split(";");
-            var answers = arr[0].split(",");
+MultipleChoice.prototype.repopulateFromStorage = function (data, status, whatever) {
+    if (data !== "") {
+        var dataEval = JSON.parse(data);
```
bnmnetp: This is strange, because in my setup data is already pre-parsed when it comes from a jquery request.  data has a timestamp an answer and a correct attribute.  Trying to JSON.parse it causes an error.

runestone/assess/js/mchoice.js
```diff
@@ -284,64 +273,109 @@ MultipleChoice.prototype.randomizeAnswers = function () {
...
+        if (this.shouldUseServer(dataEval)) {
+            var answers;
+            if (this.multipleanswers) {
+                answers = dataEval.answer.split(",");
+            } else {
+                answers = [dataEval.answer.charCodeAt(0) - 97];   // Get index for lowercase letter
```
bnmnetp: It also appears that answer and correct are switched around, as in my test answer is 'T' and correct is 'c'  this is switched around from how they are in the database.

flysaiah:
That's interesting. I'll take a look at it later tonight/tomorrow when I'm in a place with wifi (I'm on my phone). I know that I changed the way that data is passed into the database, so I can see that if your data follows the old format, it wouldn't work. I didn't think about this backwards compatibility problem.

bnmnetp:
Its probably safest to have the select statement select the columns explicitly and in the order you need them rather than using ``select *``

flysaiah:
I made the changes to how querying is done (commit added to the server repo). I'm having trouble understanding how/why your data is pre-parsed after a JSON request. Would you be able to send me some details about your test environment? I'm also unsure about how answer is "T" and correct is "c"--it seems pretty clear in ajax.py (in the server repo) that it "answer" should correspond to the database "answer." I did just change how querying is done, though, to make it safer as you said, so if that was messing up the order, it might be fixed now. Let me know if it's still an issue.

bnmnetp:
The changes to the queries look great.

I'm using version 2.5.1 of web2py, and the jquery.min version that is part of runestone 1.7.2  I'm curious what version of web2py you are testing with.

I think it may be more web2py related... If an ajax response sets the header so that the content type is correctly specified as application/json then the data passed to the success function is pre-parsed.

I ran into this with Paul R. as well when we were working on another feature.  I think the best thing to do is to use jquery.getJSON  and remove the call to JSON.parse.

flysaiah:
I'll make those changes to the files and get them in a commit--I'm using version 2.14.6 of web2py. I'm unable to find a different version, as the source code and github repo pointed to by their website https://github.com/web2py/web2py      is currently version 2.14.6. How did you get version 2.5.1?

PR #123, opened 2016-04-04
New option for timed exams and paired distractors for Parsons problems

ericsonga:
I added a notimer option on the timed exams to allow for no timer to be displayed.  Jeff Rick added paired distractors that are displayed either above or below the correct code to make it easy for the user to pick between the correct code and the distractor.

### DIFF
--- a/runestone/assess/js/timed.js
+++ b/runestone/assess/js/timed.js
@@ -50,6 +50,10 @@ Timed.prototype.init = function (opts) {
     if ($(this.origElem).is("[data-random]")) {
         this.random = true;
     }
+    this.showTimer = true; 
+    if ($(this.origElem).is("[data-no-timer]")) {
+        this.showTimer = false;
+    }
 
     this.running = 0;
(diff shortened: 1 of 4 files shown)


bnmnetp:
Barb,

Is this ready for review or is work ongoing?

One question:  with parsons.js, I'm always a bit worried about us changing the parsons.js file.  This makes it much harder to get a new version of parsons from Ville with his changes and new features.  If these are bug fixes to parsons we should make a PR back to the original to stay in sync. If these are custom for us, then is there a way to factor them out of the parsons.js source?

ericsonga:
Yes, this is ready for review.  I am not worried about staying with Ville's version.  I don't think they are working much on it and we are basically rewriting it to add the functionality I want.

bnmnetp:
@ericsonga When I accepted this PR and rebuilt my tests, I am now seeing that we are losing the indentation the multiline parson's blocks!

PR #124, opened 2016-04-12
Parsons updates

jochenrick:
This adds the following to Parsons problems: (1) You can have paired distractors by putting #paired in the RST file for the distractor line. It will be paired with the last valid line of code. (2) You can put the :order option in to specify the order that the code will be presented (e.g., :order = 3,2,0,1). (3) It does some refinements to the :maxdist option, allowing the author to specify the maximum number of distractors.

### DIFF
--- a/runestone/parsons/css/parsons.css
+++ b/runestone/parsons/css/parsons.css
@@ -11,6 +11,7 @@
 #sortableTrash { width: 38%; }
 #sortable { width: 56%; }
 .sortable-code ul {
+	position: relative;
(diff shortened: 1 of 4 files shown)


bnmnetp:
There are some merge conflicts now that I have accepted Barb's PR.

Also, can you please adjust your editor so that it uses spaces instead of tabs.  All of the code uses only spaces.  In particular your PR causes parsons.py to crash under python3 due to the mixing of tabs and spaces.

Thanks.

PR #126, opened 2016-04-14
Remove error div when running JavaScript.

jdeisenberg:
If you have a syntax error in JavaScript, the error/description/fix div appears when you click Run. If you then correct the error and click Run again, the error div remains on the screen. This pull request makes the error div disappear when you run code again.


bnmnetp:
Thanks!

Issue #113, opened 2016-01-11
Parsons problems with more than one statement in a block are indented wrong


ericsonga:
Since incorporating the latest Parsons stuff we have a problem with the indention on Parsons problems when the block of code contains multiple statements and should be indented.  Only the first line is correct (indention removed) and the rest of the code statements in the block are indented.

bnmnetp:
Barb,

Can you point me at an example?  Or copy paste the parsons example source here?  The test that I have that has code blocks seems to work fine.

Brad

ericsonga:
Here is one of the examples.  Note that the leading white space should be removed on g = p.getGreen() and on b = p.getBlue() but it is only removed on the first line.

Barb

.. parsonsprob:: Image_Decrease_GB

   Another way to get a similar effect to increasing the red, is to decrease the green and blue.  Figure out how to do that in the program above and then use that information to drag the code blocks below from the left to the right in the correct order with the correct indention. 
   -----
   from image import *
   =====
   img = Image("beach.jpg")
   =====
   pixels = img.getPixels()
   for p in pixels:
   =====
       g = p.getGreen()
       b = p.getBlue()
   =====       
       p.setGreen(g * 0.75)
       p.setBlue(b * 0.75)
   =====
       img.updatePixel(p)
   =====
   win = ImageWin(img.getWidth(),img.getHeight())
   img.draw(win)

ericsonga:
When I pasted the code it didn't keep the indention.  Here is a picture.
![screen shot 2016-01-27 at 2 06 06 pm](https://cloud.githubusercontent.com/assets/1844609/12624873/3ba17e30-c4ff-11e5-85ac-4fec06dffab2.png)

ericsonga:
I found the solution for this and it was added to my open pull request.

PR #129, opened 2016-04-29
Added timed support for ignorable elements: activecode and short answer

flysaiah:
"Ignorable" timed components are now allowed in timed, which means that any handgraded components such as short answer or activecode can now be part of a timed exam.


flysaiah:
I just found a few bugs with parsons and activecode with respect to local storage--I will reopen this when I fix them.

flysaiah:
Due to the fact that parsons evaluation involves selecting DOM elements via their ID's, parsons currently can't be made into a timed component as timed assessments hide each component when it is not the current question. Because of this, parsons only evaluated correctly when it is on-screen during evaluation, so the parsons support has been reverted in this PR.

bnmnetp:
@flysaiah, can you pull the latest and fix the conflicts.

But before you do, can you verify if you are seeing the problem with parsons problems where a multi-line block is not getting indented correctly in the rendered parsons problem.  It seems like we fix this problem but it keeps on coming back.

Brad

flysaiah:
I just tested the multiline parsons problems using the one in #113 as an example, and everything seems to work fine.

bnmnetp:
Isaiah,  I grabbed your PR into my test environment, but when I build, I get this:

screenshot 2016-05-12 16 23 55

The only thing I can think of is that it is related to not logging in.  Refresh my memory, does the rendering of the timed exam stuff rely on a login event?

flysaiah:
It does--I'm unable to replicate your problem for some reason, but I'm fairly certain it's because the runestone:login-complete isn't getting triggered in bookfuncs.js. I added some triggers that cover the missing cases that I see, which I think should fix the problem you're seeing. If that doesn't work, please send me some details about your test environment so I can replicate the problem and more precisely address it.

PR #134, opened 2016-06-02
Fixing indentation bug in parsons problems

millro04:
Split each codeblock into separate lines, storing them in an array for each codeblock (maintaining any leading spaces that were already there). We then checked to see if all the lines had the same indentation, and if they don't, we will keep the original spacing. Otherwise, we remove all the leading spaces to make it look uniform.

### DIFF
--- a/runestone/parsons/js/parsons.js
+++ b/runestone/parsons/js/parsons.js
@@ -820,8 +820,12 @@
       // Consecutive lines to be dragged as a single block of code have strings "\\n" to
       // represent newlines => replace them with actual new line characters "\n"
       //codestring = codestring.replace(/\\n\s+/g,"\\n"); // remove leading spaced if more than one line in a code block - added in below to not change the codestring
-      this.code = code.replace(trimRegexp, "$1").replace(/\\n\s+/g,"\\n").replace(/\\n+/g,"\n");
+      this.code = code.replace(trimRegexp, "$1").replace(/\\n+/g,"\n");
       this.indent = codestring.length - codestring.replace(/^\s+/, "").length;
+      alert("Code");
+      alert(this.code);
+      alert("Codestring");
+      alert(codestring);
     }
   };
   ParsonsCodeline.prototype.elem = function() {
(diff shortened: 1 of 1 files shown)


bnmnetp:
Thanks guys.

Could you paste in your test cases you used here as a comment.

You can use three backward quotes (\`\`\`) at the start and three backward quotes at the end so that github will treat it like code and not mess up indentation.  The three backquotes should be on a line by themselves.

bnmnetp:
OK, when I put the blocks in place it tells me that the first block is not indented correctly.  Are you seeing the same?

The other thing is, and this may not even be related to your changes, the characters on the line ``w = 10`` don't really line up with the line above it.

```
x = 0
=====
for i in range(10):
     x = x + 1
     message = "Indent"
=====
    w = 10
    q = 15
    t = 'Test'
   =====
    for x in range(5):
        x = 2
        y = 5
=====
```

@ericsonga do you have a couple of good test cases for us?

bnmnetp:
Oh, never mind about the first point.  I had an extra space before the ``x = 0``

bnmnetp:
The other thing I noticed is that the two lines inside the second four loop are now indented too far.  It looks like they are indented 8 characters instead of 4

bnmnetp:
I also used the following code to make sure that distractors still work.  They do.

```rst
   The following has the correct code for a method called insertInOrder plus at least one extra unneeded code statement. This method should add the passed name in alphabetic order to a private list field called nameList.  Drag the needed blocks from the left into the correct order on the right. Check your solution by clicking on the <i>Check Me</i> button.  You will be told if any of the blocks are in the wrong order or if you need to remove one or more blocks.
   -----
   public void insertInOrder(String name)
   {
   =====
     int index = 0;
   =====
     while (index < nameList.size() &&
            nameList.get(index).compareTo(name) < 0)
     {
   =====
       index++;
   =====
     } // end while
   =====
     nameList.add(index,name);
   =====
   } // end method
   =====
   nameList.add(name); #distractor
```


I think this one is ready to be accepted.

PR #136, opened 2016-06-06
Added Letters to Multiple Choice Questions/Made Modals Draggable

hangde01:


hangde01:
The change in pytutor.js under /runestone/codelens allows the modal objects to be dragged around the screen, eliminating the problem of them covering information necessary to answer the question. However, there is an invisible boundary to where they can be dragged which we haven't figured out.

PR #139, opened 2016-06-06
Fix usagedirective for building book without server

aerenchyma:
The usage directive file depends upon database access. These changes, per Lauren and Paul, fix that such that you can build a book with a usage directive included without build errors.

### DIFF
--- a/runestone/usageAssignment/__init__.py
+++ b/runestone/usageAssignment/__init__.py
@@ -46,19 +46,24 @@ class usageAssignmentNode(nodes.General, nodes.Element):
 # in html, self is sphinx.writers.html.SmartyPantsHTMLTranslator
 # The node that is passed as a parameter is an instance of our node class.
 def visit_ua_node(self,node):
-    course_name = node.ua_content['course_name']
-    s = ""
-    for d in node.ua_content['chapter_data']:
-        ch_name, sub_chs = d['ch'], d['sub_chs']
-        s += '<div class="panel-heading">'
-        s += ch_name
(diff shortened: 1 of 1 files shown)


runestone/usageAssignment/__init__.py
```diff
@@ -46,19 +46,23 @@ def __init__(self,content):
...
-            s += '<li class="simple">'
-            s += '<a href = "/runestone/static/%s/%s/%s.html">%s</a>' % (course_name, ch_name, sub_ch_name, sub_ch_name)
-            s += '</li>'
-        s += '</ul>'
-        s += '</div>'
+    try:
+        course_name = node.ua_content['course_name']
+
+        s = ""
+        for d in node.ua_content['chapter_data']:
```
bnmnetp: This seems like a big try block.  Should the check be to see ``if 'chapter_data' is in  node.ua_content:`` 

The message says that this is database related, and I'm assuming that some previous call to a database would have populated the chapter_data key of ua_content?

aerenchyma: Yeah -- thanks. I'll look and make this neater. The issue is more generally that the usageAssignment directive depends upon connection to the database, which people who are building the book in isolation don't have, so the change is to allow such users to build the book anyway (even if the book itself, like ProgramsInformationPeople, includes usageAssignment directives).

aerenchyma: I think without the db access we don't want any of the resulting HTML, but there would be other ways, too. Do you think this is better?

PR #140, opened 2016-06-07
Multiple choice feedback optional

aerenchyma:
This should make feedback, e.g. ```:feedback_a:```, etc, optional for multiple choice problems, where it did not parse correctly before.


aerenchyma:
I miscommitted on a branch. I'll fix this and reopen it.

PR #141, opened 2016-06-07
Feedback optional for multiple choice directive

aerenchyma:
This makes the feedback options, e.g. ```:feedback_a:```, actually optional instead of necessary for the build to succeed.


bnmnetp:
I accepted this, because I think they should be optional.  But I don't want it to lead to lazy question writing.  I know I've tried to write questions without feedback myself, but I've always been happy after the fact that I was forced to add good feedback for the incorrect answers.

aerenchyma:
I agree. I wonder if there's a good encouragement to add here -- a warning but not a build error? But most people don't see those when building a book anyway, I don't think. Perhaps that is not worth it (and the directive should just have quality examples).

PR #178, opened 2016-07-11
Reset and CheckMe Buttons Bug

jochenrick:
This fixes a tiny bug that stopped Reset and CheckMe buttons (in Parsons) from working in Firefox.


bnmnetp:
Thanks @jochenrick Could you take a look at the selenium integration tests I wrote for activecode -- mostly as an example -- and do something similar for Parsons?  the tests are in runestone/activecode/tests

jochenrick:
Sure, Brad. I'll put that onto my todo list.

PR #195, opened 2016-07-15
when building the book, populate/update the questions table.

bnmnetp:
@RunestoneInteractive/contributors 

I have tested this on the overview book and it populates the questions table quite nicely when you build.   IF the DBXXXX environment variables are defined it will attempt to update the database whenever the book is rebuilt. 

When the book is rebuilt, if a divid is already in the database, then the text of the question is compared.  If the question has changed it is updated.  If no divid is found then the question is inserted.

We are probably going to want to create a new directive like this:

```
.. component_from_db:: divid
```

This will simply get the text for the question from the database and then as Sphinx to recursively parse the inserted question.  I *think* it should work 😄    This would be the logical way to easily generate an assignment page from the instructors interface.

### DIFF
--- a/runestone/activecode/activecode.py
+++ b/runestone/activecode/activecode.py
@@ -22,6 +22,7 @@ from docutils.parsers.rst import directives
 from docutils.parsers.rst import Directive
 from .textfield import *
 from sqlalchemy import create_engine, Table, MetaData, select, delete
+from runestone.server.componentdb import addQuestionToDB
 
 try:
     from html import escape  # py3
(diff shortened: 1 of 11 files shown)


bnmnetp:
Digging in to this raises some additional questions.

For example we have pip and thinkcspy  There are probably some divids in common between those two books, but since they have diverged the questions may have changed substantially...

Right now we have defined the table so that the divid must be unique in the table.  I think we want to keep that, rather than have basecourse+divid be unique.  But... this may result in some work in updating divids in various books.

I've long wanted to bring some sanity to the divid naming scheme, maybe now that we have a database table containing all of them we can think about a way to do so...

Issue #206, opened 2016-07-29
Need to create an exercise directive or allow for question text to be added to an activecode


bnmnetp:
This problem revealed itself in testing the new assignment builder.  The problem is that you can search for ex_2_5 and you will get the `.. actex` directive.  But when when we go to build a file of questions for this assignment, we have no way to know what the actual question is, because the question has always been outside of the directive itself.

bnmnetp:
@RunestoneInteractive/contributors I would like to solicit some opinions / ideas / dialog on how we should solve this one.

We cannot make it mandatory to include some question text as part of the directive as active codes are used for exercises and examples alike.

We could modify the `..actex` directive to include some description. as this is already used for the exercises and is implemented as a subclass of the ActiveCode class.

I think doing it as an option is quite problematic as many programming exercise descriptions are fairly long, and may contain images or multi-line code snippets.

A `.. exercise::` directive would give us a lot of control over the formatting of the exercise.

bnmnetp:
I am leaning strongly to the following:

```
.. question:: draw_turtle_flower

   Using turtle graphics, draw an image the looks like the following:
   
   .. image:: path/to/flower.png

   .. actex:: ex_5_4
      :language: python

```

This will preserve the activecode id for all existing questions.

But when the page containing the assignment is generated we'll have both the question text and the activecode component we need together.

presnick:
This looks but three additional things will have to be done:

1. When processing the .. question directive, may need to store stuff in the database about subcomponents, like the actex ex_5_4.
2. When responding to assignments/problem ajax call, will need to provide that info from the database, like we currently do for activecodes (included files, prefix code, etc.)
3. The javascript that handles the response to assignments/problem will have to handle the additional information, and not just assume that it's getting just an activecode back.

presnick:
Maybe if we could intercept the HTML that sphinx produces for the question directive, it would be great to capture it and store it in the DB.

aerenchyma:
I also like that idea with preservation of the activecode id ^, and I agree that it would be good to associate all of the subcomponent ids (during build, presumably?). It seems like there might be a way of doing that during the build that doesn't require intercepting the sphinx-produced HTML, but perhaps that's best.

bnmnetp:
I think this question directive is problematic for grading in a few ways.

[...]

PR #210, opened 2016-08-03
Runestone directive

bnmnetp:
This PR creates a parent class  RunestoneDirective

It adds an option spec containing author, difficulty, and tags that can be processed by all RunestoneComponents.

It also splits the fitb directive into its own folder.


bnmnetp:
This PR now contains the question directive as outlined in #206  

The `.. question:: uniqued` directive replaces the #. numbered lists in the exercises at the end of each chapter.

The questions are autonumbered

The questions are *nearly* full width.

screenshot 2016-08-04 15 41 35

PR #253, opened 2016-08-29
Add files for external exercise directive.

aerenchyma:
This commit creates an external exercise directive as previously discussed on Slack, allowing an author to specify an exercise external to a Runestone textbook but still have a tie to the Runestone grading interface and a div_id. Format is as follows:

`.. external:: unique_div_id_here`

      content of question, perhaps with links, here

Excepting that you can't do anything in the textbook and therefore can't autograde, it's the same as all other gradeable directives (in fact it should do essentially what all gradeable directives should do. There's a note in the code about that). It gets added to the db with type 'external'.


aerenchyma:
I'm testing this again today and it's not parsing RST the way I'd like it to. I am figuring out why.

PR #257, opened 2016-08-31
Autograde

presnick:
Adds optional autograde parameter for all Runestone directives.  In addQuestionToDB, writes that parameter into the questions DB.

Also adds gradebutton flag for activecode.


presnick:
Note: rqeuires a RunestoneServer change to add autograde parameter to questions table in the DB.

bnmnetp:
@presnick I left you comments on the server change that need to be addressed before I can accept these together.

presnick:
@bnmnetp I think the problems on the server change have been addressed. Please let me know if there's anything pending that I'm not aware of.

PR #370, opened 2017-02-13
Enclosed data-include in quotes to fix multiple includes

FranciscoMoya:
Currently activecode is unable to include multiple snippets from other activecode  components. Value of the data-include attribute is inserted without quoting the identifiers and using a space separator.  Therefore second and later included identifiers are seen as textarea attributes instead of part of the data-include attribute.


bnmnetp:
Thanks!  

I've got a few small PR's to test this weekend, including this one.

PR #371, opened 2017-02-16
Popup warning when leaving a timed test in progress

BunsenMcDubbs:
Now, when a user tries to navigate away from a timed test, the browser stop them with a popup and warn the user. Helps prevent accidental navigation to other pages. Covers all possible navigation attempts ("back" button, clicking links, url change).


bnmnetp:
Thanks!  I'll  look at this more closely over the weekend.

bnmnetp:
Great idea.  This will help a lot.

PR #373, opened 2017-02-28
Fix timed exam pagination accessibility by screen reader and tab focus

RSLi:
Originally timed exam pagination cannot be focused by tab and screen reader, now they are fully accessible.


RSLi:
The reason of the existing problem and the solution is explained [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role)

jochenrick:
Looks good.

PR #382, opened 2017-04-25
Data Files with Java

CABowers:
Enabled supplemental files to be run alongside Java activecode. 
Have working examples here on the Images/imageBasics.html page.
https://github.com/CABowers/APCSAReview


runestone/activecode/js/activecode.js
```diff
@@ -1844,3 +2095,207 @@ $(document).bind("runestone:logout",function() {
...
+    string = Utf8Encode(string);
+
+    x = ConvertToWordArray(string);
+
+    a = 0x67452301; b = 0xEFCDAB89; c = 0x98BADCFE; d = 0x10325476;
+
+    for (k=0;k<x.length;k+=16) {
+        AA=a; BB=b; CC=c; DD=d;
+
+        a=FF(a,b,c,d,x[k+0], S11,0xD76AA478);
```
bnmnetp: Sorry, this is a bit incomprehensible.  What is long block of code supposed to be doing?  I don't think anyone could ever figure out what this is for much less fix it or debug it if it was somehow broken.

bnmnetp: OK, I see now that it is part of a larger implementation of MD5.  But you should not copy and paste the code into activecode.js.    You should add a script tag to link to the implementation so we can keep up to date with their bug fixes.  At a minimum you should grab the original .js file and add it separately to our repository.

CABowers:
Thanks for the info, I believe I have done what you asked. Let me know if I need to make anymore changes.

runestone/activecode/js/MD5.js
```diff
@@ -0,0 +1,203 @@
+/**
```
bnmnetp: OK, so my last comment is simply that if you copied this from somewhere else that we can publicly link to, lets just do that.  There's no reason to have a copy that can get out of date when someone else fixes a bug in the code you copied.

CABowers: How would you like me to publicly link an outside source to this repository. I am not quite sure how to do that.

bnmnetp: You just use the full http://..... form of the URL to link to it rather than a relative link.

CABowers:
Thanks for working with me on this. Let me know if there is anything else you want me to update.

CABowers:
I forgot to test again after merging from the upstream. Merging with the updated Runestone is causing an error on line 22 of activecode.js. I will look into it.

CABowers:
My commits should be good to go. On my computer Chrome was having issues with the following line of code:
`RunestoneBase.prototype.init.apply(this, arguments); // ActiveCode ln:22`
I tested this on a build of Runestone without my changes and it persisted. When I tested on Firefox the issue was gone. There must be an issue with how my version of Chrome is interpreting the JavaScript, which is unrelated with my commit.

PR #383, opened 2017-04-26
Audio Tour Revamp

PC98:
1. Changed the audio tour UI - instead of a separate screen popping up, the audio tour now launches in the same screen, right below the "Audio Tour" button.

2. Because the audio tour doesn't require a separate screen to pop up now, the user can have audio tours from multiple coding examples open simultaneously.

3. Earlier, images were being used as place-holders for media control buttons (that is, the "play button", "next-audio button" etc. were images); I have replaced them with actual HTML button tags.

4. A bunch of bug fixes, for example, earlier audio would play when the user clicked on the "next audio" button even though "pause audio" was selected.

I have tested the new audio tour on Chrome, Safari and Firefox. I also ran the changes through Barbara and Jeff.


ericsonga:
I tested with the first active code at 
http://interactivepython.org/runestone/static/StudentCSP/CSPrinTeasers/computeNumbers.html 

I got a "Failed to load resource.  The server responded with a status of 403 (forbidden).  I bet it is because it uses http and not https to get the audio file.

PC98:
Hey Barbara,
I am actually getting a 404. It seems the audio files no longer exists at the URLs they used to. activecode.js loads audio files from URLs that look like:
http://media.interactivepython.org/[eBook Name]/audio/[key].[file extension], one such example is:

http://media.interactivepython.org/StudentCSP/audio/BMI-line1.mp3

I tried accessing the same URL using the HTTPS protocol but that didn't work either. Do you know if the host or path for the audio files have been changed?

bnmnetp:
I just checked, and two things:

 /StudentCSP/audio/BMI-line1.mp3

* The folder for these has always been studentcsp 
* There is no BMI-line1.mp3 file there,  There is a BMI-line1.wav file.

Brad

PC98:
Ahh I see what the issue is, thanks for pointing it out Brad. I have made a small change and it fixed the error for me. I will close this PR and make a new one.

PR #389, opened 2017-06-12
Improve multiple-choice questions

bjones1:
This PR contains three changes. The first two are fairly minor:

1. A bit of cleanup -- removing unused imports.
2. A fix to assume that multiple-choice questions whose `:correct:` field contains a list is a multiple-answer question.

The third is the primary change:

3. Allow block or inline markup as part of the question and feedback text by optionally specifying these as a bulleted list.

I use [CodeChat](https://pythonhosted.org/CodeChat/README.html), which allows me to render reStructuredText in comments nicely, hence some of the formatting in comments.

Feel free to request changes/edits/whatever, given that this is my first PR to this project. I look forward to meeting everyone in person next week!

### DIFF
--- a/runestone/assess/multiplechoice.py
+++ b/runestone/assess/multiplechoice.py
@@ -17,10 +17,7 @@ __author__ = 'bmiller'
 
 from docutils import nodes
 from docutils.parsers.rst import directives
-from docutils.parsers.rst import Directive
 from .assessbase import *
-import json
-import random
 from runestone.server.componentdb import addQuestionToDB, addHTMLToDB
 
 
@@ -45,7 +42,8 @@ def visit_mc_node(self,node):
         node.mc_options['random'] = 'data-random'
     else:
         node.mc_options['random'] = ''
-    if 'multiple_answers' in node.mc_options:
+    # Use multiple_answers behavior if explicitly required or if multiple correct answers were provided.
+    if ('multiple_answers' in node.mc_options) or (',' in node.mc_options['correct']):
         node.mc_options['multipleAnswers'] = 'true'
     else:
         node.mc_options['multipleAnswers'] = 'false'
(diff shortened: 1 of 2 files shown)


runestone/assess/multiplechoice.py
```diff
@@ -93,22 +91,46 @@ def depart_mc_node(self,node):
...
+            -   Feedback for answer A.
+
+                This may also span multiple paragraphs and include any markup.
+                However, there can be only one item in this unordered list.
+
+        - Text for answer B.
+
+            -   Feedback for answer B.
+
+        ... and so on.
```
bnmnetp: This is nice, and a real improvement for easier writing!

bjones1:
Thanks! The HTML should be very similar (though there are some differences in whitespace). I'll check on the timed exam and let you know.

On a side note, I'm also thinking of doing something similar for fill in the blank problems using [definition lists](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#definition-lists) -- does that sound reasonable?

bnmnetp:
Yes!!  The fill in the blank syntax is really awkward.   I had an idea to rewrite it using an inline `role` for the blanks.  Lets talk about the pro's and cons of definition lists versus a role.

bjones1:
Hmm...what about something like:

``` rst
.. fillintheblank:: unique_id_1034

    Complete the sentence: :blank:`first_blank` had a :blank:`second_blank` lamb.

    .. blank_answer:: first_blank
        :Mary: Was the correct answer!
        :.*: Wrong!

    .. blank_answer:: second_blank
        :little: That's right!
        :.*: Nope.
```

I'd like if we could insert empty roles, such as `:blank:`, but this is a field list. Or, perhaps `blank_answer` with no arguments would assume sequential numbering?

bjones1:
Testing with the timed exam feature shows no problems. (It would be good to include info on it at http://runestoneinteractive.org/build/html/directives.html -- timed questions aren't discussed there.)

bjones1:
Hmmm. After looking at the existing code, how about:

``` rst
.. fillintheblank:: unique_id_1034

    blank:: blank_id_11

        Complete the sentence:

        :Mary: Was the correct answer!
        :.*: Wrong!

    blank:: blank_id_xxx

        had a (your answer) lamb.

        :little: That's right!
        :.*: Nope.

bnmnetp:
I liked your previous one a lot more.  I think we need to make writing the questions as easy as possible.  That is why I dislike the current syntax for fill in the blank.  Its impossible to remember and get right.  You suggestion from 3 days ago is really easy to remember and write.

bjones1:
I agree. I'll work toward a larger rewrite, which will involve modifying the JavaScript as well. The existing JS takes HTML and transforms it into an interactive question. I think it's easier to write the HTML as an interactive question -- any concerns about that approach?

bnmnetp:
We can whiteboard this when you are here.   My goal has since version 2 a couple years ago was to have the html be pretty simple so that in theory someone could use runestone by writing html and including the correct js files.    I'm not sure we have achieved that goal in practice but I would not want to go "back" too far in having loads of complex html generated by Sphinx.

PR #393, opened 2017-06-20
Add: Make the keep_warning Sphinx config value True by default.

bjones1:


bnmnetp:
I'm not sure I want to have these warnings appear in student facing books.  I can see the advantage in preview mode but in general???

PR #399, opened 2017-06-22
Added functionality to reset timed exams.

conzty01:
This is still a 'work in progress'.  What needs to be added is a way for the javascript to check with the server and see if the exam has been taken before it will activate the reset button.  It also needs to record that the exam has been reset.


conzty01:
This is now functional.

runestone/assess/js/timed.js
```diff
@@ -456,6 +468,32 @@ Timed.prototype.pauseAssessment = function () {
     }
 };
 
+Timed.prototype.resetAssessment = function () {
+    if (this.taken) {
+        if (window.confirm("Only reset the exam if you encountered problems completing it. The Instructor will be notified of the reset.")) {
+            eventInfo = {"event":"timedExam","act":"reset","div_id":this.divid,
+                         "course":eBookConfig.course,"correct":this.score,"incorrect":this.incorrect,
+                         "skipped":this.skipped,"time":this.timeTaken,"reset":True}
```
bnmnetp: the `"reset":True` should be `"reset":true`   Capital True is Python.

conzty01:
Did the page refresh after you hit the reset button? And, if so, were you able to retake the quiz and submit answers?

bnmnetp:
The page did refresh, but I was not able to answer any of the unanswered questions.

conzty01:
Fixes on the [server-side ](https://github.com/RunestoneInteractive/RunestoneServer/pull/958)

bnmnetp:
@conzty01 

I'm still trying to get this to work the way I think it should.  Maybe we didn't have a good understanding on this one.

1.  I start the exam and answer a couple of questions.
2. I force close the page.  -- nothing in the database yet
3. I go back to the page.  Now I can see the questions.  None of them have the feedback showing, but I cannot answer any questions.  The timer is not running.  The Reset button is active now.  -- this seems like a bad state.  Now I appear to have as much time as I want to look at and think about these questions.
4. I click reset and OK
5. The page refreshes and all of the feedback is showing for all of the questions.  I still can't answer any.

-------
What I would expect to happen

1. I start the exam and answer some questions.
2. I force close the page.
3. I go back to the page.  I can see the feedback for the questions I already answered before the force close, but I cannot see feedback for unanswered questions.   Reset is available.
4. I click on reset.  -- a database entry is made indicating the reset, and the questions I had already answered.
5. Feedback is still there for the questions I already answered. But now I can answer the questions that I had not answered before.
6. I submit my exam -- another database row is added with my final information.

conzty01:
Our understandings of what you were looking for were a bit off, but I have not been able to replicate the issue you are having with not being able to answer the questions after the reset.  I am *still* working on getting my VM back up and running as I am having issues installing, but I will start to work on implementing the behavior you are looking for when I can.

bnmnetp:
OK, please describe for me your understanding of what I should expect.  Once we have compared notes we can decide on the best course.

conzty01:
1. Start the exam and answer some questions.
2. Force close the page.
3. Reload page. See feedback for the questions previously answered before the force close, but not for unanswered questions. Reset is enabled.
4. Click on reset. Database is notified of reset and previous answers.
5. Feedback is still there for the questions already answered. But now the user can answer the questions that that were not answered before.
6. Submit the exam. Feedback is displayed for entire exam and database is notified of completed exam.

conzty01:
Might I suggest, that feedback would be kept hidden until the entire exam is completed. This would eliminate the possibility of being able to read the question without the timer running and the lack of the green and red highlighting would be an indication to a student browsing by or double checking the submission that the exam was not finished properly.

Such a design might behave like:

1. Start exam and answer some, but not all, questions
2. Force navigation away from the page
3. Reload page and see that exam appear similar to how it was before starting but with an updated timer to reflect how much time they took previously.
4a. Hit 'Reset' to reset entire exam
5a. Page reloads with a fresh copy of the exam and record of reset is sent to the database.

4b. Hit start exam again and resume the exam as before but without the ability to edit previously answered questions.
5b. On completion, the exam displays all feedback.

bnmnetp:
@conzty01, I like your second option a lot.

conzty01:
So, my newly installed VM doesn't have the same functionality with the reset feature that it did before the crash, so I believe we are in the same boat now.  Because of the lack of functionality and the new approach that you would like from my last comment, I have gone through the `timed.js` script, excluding any changes I made for the reset, to try to figure out where/when might be the best place to check the localStorage and server for submitted exams. The "flow" chart that I found is below.

conzty01:
The bulk of the loading of the timed exam is in done in the `tookTimedExam( )` function so I figure that that would be the most appropriate place to check for previous answers. However, I am stuck on how to proceed with checking-- i.e. which place do I check for answers first and when in the order should it occur.  My initial idea is to check right at the beginning of `tookTimedExam( )` but from there I am unsure. Thoughts, @bnmnetp ? The flowchart begins from the "Timedexam init" box in the top left corner.

bnmnetp:
So, if there is not a full timed_exam entry in localStorage but there are individual entries for one or more of the timed elements that is a really strong indication that the exam was interrupted or force quit somehow right?

I think you will already want to invent a local storage way to keep track of the time.  So maybe you could also use local storage to keep track of whether the exam was submitted or not??

bnmnetp:
@conzty01 Not sure my response made sense, and I'm not sure I understand your flowchart.   Maybe we should do a hangout tomorrow morning to really work this out.

conzty01:
- [ ] Re-order `timed.js` to eliminate extraneous/outdated code.
- [ ] Create timestamp for when exam was started in localStorage.
- [ ] Keep track of time taken on the exam in localStorage.
- [ ] If Database has no results for exam, check localStorage for answers.
  - [ ] If the difference between the timestamp and current time is greater than the length allotted for the exam, do not allow students to continue the exam.
  - [ ] If the difference listed previously is less than the allotted time, allow the students to continue the exam without changing previous answers.
- [ ] Display feedback at the end of an exam only if the exam is either finished, or there is a database entry.
- [ ] Allow exam to be reset if the Database has no record of an exam and localStorage says that all questions were skipped.

conzty01:
[Server Updates](https://github.com/RunestoneInteractive/RunestoneServer/pull/965)

conzty01:
I've updated the reset function if you'd like to take a look, @bnmnetp , but I am still working on being able to resume an incomplete exam.

bnmnetp:
It seems to me that this is a good point to stop with this one.   I can see some benefit to being able to restore answers prior to some kind of catastrophic end to the exam, but the additional complexity it would add to all of the components and the already fragile nature of the timed_exams make me think that a simple reset capability is all that is needed.

runestone/common/js/runestonebase.js
```diff
@@ -1,35 +1,7 @@
-/**
```
bnmnetp: This should not have been deleted.  Any idea why it was?

runestone/common/js/runestonebase.js
```diff
@@ -1,35 +1,7 @@
...
-  *
-  * 4. provide a Selenium based unit test
-  *
- **/
-
 function RunestoneBase () {   // Basic parent stuff
 
 }
 
-RunestoneBase.prototype.init = function(opts) {
```
bnmnetp: this should not have been deleted either

runestone/common/js/runestonebase.js
```diff
@@ -56,14 +28,11 @@ RunestoneBase.prototype.logRunEvent = function (eventInfo) {
 
 RunestoneBase.prototype.checkServer = function (eventInfo) {
     // Check if the server has stored answer
-    if (this.useRunestoneServices || this.graderactive) {
```
bnmnetp: keep this

runestone/common/js/runestonebase.js
```diff
@@ -56,14 +28,11 @@ RunestoneBase.prototype.logRunEvent = function (eventInfo) {
...
-        let data = {};
+    if (this.useRunestoneServices) {
+        var data = {};
         data.div_id = this.divid;
         data.course = eBookConfig.course;
         data.event = eventInfo;
-        if (this.sid) {
```
bnmnetp: keep this

runestone/assess/js/timed.js
```diff
@@ -318,14 +326,9 @@ Timed.prototype.renderFeedbackContainer = function () {
 Timed.prototype.createRenderedQuestionArray = function () {
     // this finds all the assess questions in this timed assessment and calls their constructor method
     // Also adds them to this.renderedQuestionArray
-    // todo:  This needs to be updated to account for the runestone div wrapper.
     for (var i = 0; i < this.newChildren.length; i++) {
         var tmpChild = this.newChildren[i];
         opts = {'orig':tmpChild, 'useRunestoneServices':eBookConfig.useRunestoneServices}
-        if ($(tmpChild).children("[data-component]")) {
```
bnmnetp: these should not have disappeared either

PR #402, opened 2017-06-23
Add no longer ignored conf.py for the fitb tests so they will run

aerenchyma:


aerenchyma:
Let me know if the merge commit (just to ensure no conflicts) causes a problem and I'll fix it. I don't think it will, this should be master + my conf.py

PR #404, opened 2017-06-27
Fix: Use a single DBURL environment variable...

bjones1:
instead of individual DBUSER/DBPASS/DBHOST/DBNAME. This makes setup a bit easier, and allows supporting other databases. It also cleans cruft by using get_dburl, rather than custom code in many places.


bjones1:
Makes sense. Are these scripts in github somewhere? If so, would you like any help with fixes/upgrades for them?

bjones1:
Awesome! Thanks!

PR #408, opened 2017-06-30
fixed bug that prevented correct parsons problems from being added to parsons_answers

gardhi01:


runestone/parsons/js/parsons.js
```diff
@@ -126,7 +126,7 @@ LineBasedGrader.prototype.grade = function() {
 			       feedbackArea.html("Perfect!  It took you only one try to solve this.  Great job!");
 			    }
 				correct = true;
-				problem.helpButton.disabled = false;  // bje
+				// problem.helpButton.disabled = false;  // bje
```
bnmnetp: Really?  This is the solution?  Can you give me a short explanation?

bnmnetp: I glanced at the code but I don't see the connection

bnmnetp: @ericsonga, Hillary is correct when I try to check an answer that is correct, this error stops the chain, and then logging does not happen for correct answers!  I wonder if you really meant to say this.helpButton.disabled = false; ???

bnmnetp:
I think this was only failing for non-adaptive parsons, so I put in a check to see if there was a help button before disabling it.

PR #409, opened 2017-07-01
Removed line that wasn't needed

ericsonga:
Removed the disabling of the help button.  It should have been inside a
conditional and only executed if the problem was adaptive.  But, now
that I have made the help button always active it isn’t needed.


ericsonga:
I see that you added the conditional, Brad, but we can just remove that line.

ericsonga:
I also found where the size of the Parsons was being limited by the addition of a div with a class of runestone which is limited to 500 px wide.  I removed that to allow the Parsons problems to take up more width.

bnmnetp:
I fixed the timed exam problem introduced by the runestone wrapper, and we can fix the parsons width using CSS.  The benefits of the wrapper are worth the effort.  I'll remove those lines from parsons.js

Issue #405, opened 2017-06-27
Fill in the blank ideas


bjones1:
Recapping the discussion from #389, my proposed format was:

``` rst
.. fillintheblank:: unique_id_1034

    blank:: blank_id_11

        Complete the sentence:

        :Mary: Was the correct answer!
        :.*: Wrong!

    blank:: blank_id_xxx

        had a (your answer) lamb.

        :little: That's right!
        :.*: Nope.
```

In the past, this was generated from HTML such as:

``` html
<div class="runestone">
<p data-component="fillintheblank" data-casei="false" id="fill1512">
    
<span data-blank>
    What is the value of 25 expressed as an octal number (base 8)?
<span data-feedback="regex" style="display: none" id="blank21_feedback1">\b25\b</span>
<span data-feedback="text" style="display: none" for="blank21_feedback1">That's base 10!</span>
    
<span data-feedback="regex" style="display: none" id="blank21_feedback2">.*</span>
<span data-feedback="text" style="display: none" for="blank21_feedback2">25 in octal please!</span>
    
<span data-answer style="display: none" id="blank21_answer">\\b31\\b</span>
</span>

<span data-blank>
    Second blank.
<span data-feedback="regex" style="display: none" id="blank22_feedback1">\bxx\b</span>
<span data-feedback="text" style="display: none" for="blank22_feedback1">That's base 10!</span>
    
<span data-feedback="regex" style="display: none" id="blank22_feedback2">.*</span>
<span data-feedback="text" style="display: none" for="blank22_feedback2">xx in octal please!</span>
    
<span data-answer style="display: none" id="blank22_answer">\\b22\\b</span>
</span>

</p>
</div>
```

The purpose, as I understand it, was to make it possible for other authors to create questions from HTML, not just from the RST. If so, why not a more compact, easily parsed format based on JSON?

``` html
<div class="runestone">
<p data-component="fillintheblank" data-casei="false" id="fill1512">

    What is the value of 25 expressed as an octal number (base 8)?
<script type="application/json">
  [
    {
      "regex" : "\\b31\\b"
      "text" : "That's <b>correct</b>."
    }, {
      "regex" : "\b25\b",
      "text" : "That's base 10!",
    }, {
      "regex" : ".*",
      "text": "25 in octal please!"
    },
  ]
</script>
</p>
</div>
```

Or, even better, make the *entire* question a JSON element that will be read and reformatted by JS.

bnmnetp:
[...]

PR #412, opened 2017-07-04
Fix: Provide a more functional, usable fill-in-the-blank question.

bjones1:
This implements the changes discussed in #405, plus adds a few new features and bug fixes. The final form:

``` rst
.. fillintheblank:: some_unique_id_here

    Complete the sentence: |blank| had a |blank| lamb. 
    One plus one is: (note that if there aren't enough blanks for the feedback given, 
    they're added to the end of the problem. So, we don't **need** to specify a blank here.)

    -   :Mary: Is the correct answer.
        :Sue: Is wrong.
        :x: Try again. (Note: the last item of feedback matches anything, regardless of the string it's given.)
    -   :little: That's right.
        :.*: Nope.
    -   :2: Right on!
        :2 1: Close.... (The second number is a tolerance, so this matches 1 or 3.)
        :x: Nope. (As earlier, this matches anything.)
```

### DIFF
--- a/runestone/common/project_template/conf.tmpl
+++ b/runestone/common/project_template/conf.tmpl
@@ -92,10 +92,20 @@ pygments_style = 'sphinx'
 
 # `keep_warnings <http://www.sphinx-doc.org/en/stable/config.html#confval-keep_warnings>`_:
 # If true, keep warnings as “system message” paragraphs in the built documents.
-# Regardless of this setting, warnings are always written to the standard error 
+# Regardless of this setting, warnings are always written to the standard error
 # stream when sphinx-build is run.
 keep_warnings = True
 
+# `rst_prolog <http://www.sphinx-doc.org/en/stable/config.html#confval-rst_prolog>`_:
+# A string of reStructuredText that will be included at the beginning of every
+# source file that is read.
+rst_prolog = (
+# For fill-in-the-blank questions, provide a convenient means to indicate a blank.
+"""
+
+.. |blank| replace:: :blank:`x`
+"""
+)
 
 # -- Options for HTML output ---------------------------------------------------
 
--- a/runestone/fitb/js/fitb.js
+++ b/runestone/fitb/js/fitb.js
@@ -32,91 +32,41 @@ FITB.prototype.init = function (opts) {
     this.useRunestoneServices = opts.useRunestoneServices;
     this.origElem = orig;
     this.divid = orig.id;
-    this.questionArray = [];
     this.correct = null;
-    this.feedbackArray = [];
-    /* this.feedbackArray is an array of array of arrays--each outside element is a blank. Each middle element is a different "incorrect" feedback
-    that is tailored for how the question is incorrectly answered. Each inside array contains 2 elements: the regular expression, then text */
-    this.children = [];   // this contains all of the child elements of the entire tag...
-    this.correctAnswerArray = [];   // This array contains the regular expressions of the correct answers
-
-    this.adoptChildren();
-    this.populateCorrectAnswerArray();
-    this.populateQuestionArray();
+    // The format of this array:
+    //
+    // .. code-block::
+    //  :number-lines:
+    //
+    //  [ // Overall data structure is an array.
+    //      [ // blank1
+    //          [ // correct answer
+    //              'regex1.1',
+    //              'text1.1'
+    //          ], [ // incorrect1
+    //              'regex1.2',
+    //              'text1.2'
+    //          ], // Followed by more incorrects.
+    //      ], [ // blank2
+    //          etc.
+    //      ]
+    //  ]
+    //
+    // Find the script tag containing JSON and parse it. See `SO <https://stackoverflow.com/questions/9320427/best-practice-for-embedding-arbitrary-json-in-the-dom>`_.
+    this.feedbackArray = JSON.parse(this.scriptSelector(this.origElem).html());
 
     this.casei = false;   // Case insensitive--boolean
     if ($(this.origElem).data("casei") === true) {
         this.casei = true;
     }
-    this.populateFeedbackArray();
     this.createFITBElement();
     this.checkServer("fillb");
 };
 
-/*====================================
-====    Functions parsing data    ====
-====   out of intermediate HTML   ====
-====================================*/
-
-FITB.prototype.adoptChildren = function () {
-    // populates this.children
-    var children = this.origElem.childNodes;
-    for (var i = 0; i < this.origElem.childNodes.length; i++) {
-        if ($(this.origElem.childNodes[i]).is("[data-blank]")) {
-            this.children.push(this.origElem.childNodes[i]);
-        }
-    }
-};
-
-FITB.prototype.populateCorrectAnswerArray = function () {
-    for (var i = 0; i < this.children.length; i++) {
-        for (var j=0; j < this.children[i].childNodes.length; j++) {
-            if ($(this.children[i].childNodes[j]).is("[data-answer]")) {
-                this.correctAnswerArray.push($([this.children[i].childNodes[j]]).text().replace(/\\\\/g,"\\"));
-            }
-        }
-    }
-};
-
-FITB.prototype.populateQuestionArray = function () {
-    for (var i = 0; i < this.children.length; i++) {
-        for (var j = 0; j < this.children[i].childNodes.length; j++) {
-            if ($(this.children[i].childNodes[j]).is("[data-answer]")) {
-                var delimiter = this.children[i].childNodes[j].outerHTML;
-
-                var fulltext = $(this.children[i]).html();
-                var temp = fulltext.split(delimiter);
-                this.questionArray.push(temp[0]);
-                break;
-            }
-        }
-    }
-};
-
-FITB.prototype.populateFeedbackArray = function () {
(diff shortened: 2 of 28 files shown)


runestone/common/project_template/conf.tmpl
```diff
@@ -92,10 +92,20 @@ pygments_style = 'sphinx'
...
-# Regardless of this setting, warnings are always written to the standard error 
+# Regardless of this setting, warnings are always written to the standard error
 # stream when sphinx-build is run.
 keep_warnings = True
 
+# `rst_prolog <http://www.sphinx-doc.org/en/stable/config.html#confval-rst_prolog>`_:
+# A string of reStructuredText that will be included at the beginning of every
+# source file that is read.
+rst_prolog = (
+# For fill-in-the-blank questions, provide a convenient means to indicate a blank.
```
bnmnetp: Existing books will need to update their conf.py in order to get this.  We'll need to do that for pip, thinkcspy, pythonds, etc.

Although if I understand this correctly an author is free to use :blank:`foo`  if they want.

bjones1: Yes, that's correct; any form of ``:blank:`foo` `` (where `foo` can be anything) works.. Would you like me to submit PRs for these books? Or would you prefer to edit them yourself?

bjones1:
Re; `test/_sources/index.rst`, will do - I'll push an extra commit with this.

bnmnetp:
ready to merge this as soon as test/.../index.rst  is updated for fitb.

Thanks!

runestone/fitb/js/fitb.js
```diff
@@ -32,91 +32,20 @@ FITB.prototype.init = function (opts) {
...
-        this.casei = true;
-    }
-    this.populateFeedbackArray();
+    // See comments in fitb.py for the format of ``feedbackArray`` (which is identical in both files).
+    //
+    // Find the script tag containing JSON and parse it. See `SO <https://stackoverflow.com/questions/9320427/best-practice-for-embedding-arbitrary-json-in-the-dom>`_.
+    this.feedbackArray = JSON.parse(this.scriptSelector(this.origElem).html());
+
     this.createFITBElement();
     this.checkServer("fillb");
```
bnmnetp: I'm creating a checklist/writing up a document on writing a component...

You should check for the existence of a sid key in the opts passed to the constructor.  If sid is populated (as it will be for grading) then checkServer will do the right thing in the grading interface with respect to populating the correct answer.

In fact the right way to do this is to add an init to `RunestoneBase` that takes care of this for everything and then make sure that each component calls  `RunestoneBase.init.apply(this, arguments)`

bjones1: Sounds good! Done in 11d9059c3e25c187224d18bf9d6266e8007944ff.

bjones1:
I'll work on updating the tests!

bnmnetp:
Great shouldn't take you long, you should only need to update index.rst in fitb/test

bjones1:
I refactored the test framework to make it cross-platform, and DRY the code. FITB tests pass when using the Chrome driver, but not under PhantomJS -- for some reason, the PhamtomJS driver doesn't transform the HTML on the page into question format (by running code in fitb.js), while Chrome does. ??? I'm not sure if this fails on other platforms or not. I've only tested under Windows.

On Windows/Chrome, I see two test failures. Do these fail on other OSes?

```
ERROR: test_history (runestone.activecode.test.test_activecode.ActiveCodeTests)
  File "E:\Runestone\RunestoneComponents\runestone\activecode\test\test_activecode.py", line 41, in test_history
    rb.click()
  File "E:\Downloads\Python36-64\lib\site-packages\selenium\webdriver\remote\webelement.py", line 77, in click
    self._execute(Command.CLICK_ELEMENT)
  File "E:\Downloads\Python36-64\lib\site-packages\selenium\webdriver\remote\webelement.py", line 493, in _execute
    return self._parent.execute(command, params)
  File "E:\Downloads\Python36-64\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 256, in execute
    self.error_handler.check_response(response)
  File "E:\Downloads\Python36-64\lib\site-packages\selenium\webdriver\remote\errorhandler.py", line 194, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.WebDriverException: Message: unknown error: Element <button class="btn btn-success run-button" disabled="disabled">...</button> is not clickable at point (334, 173). Other element would receive the click: <div class="ac_actions col-md-12">...</div>
  (Session info: chrome=59.0.3071.115)
  (Driver info: chromedriver=2.30.477700 (0057494ad8732195794a7b32078424f92a5fce41),platform=Windows NT 10.0.14393 x86_64)


FAIL: test_dnd4 (runestone.dragndrop.test.test_dragndrop.DragAndDropQuestion_Tests)
Incorrect answer changed to correct
  File "E:\Runestone\RunestoneComponents\runestone\dragndrop\test\test_dragndrop.py", line 126, in test_dnd4
    self.assertIn("alert-danger", cnamestr)
AssertionError: 'alert-danger' not found in 'alert alert-success draggable-feedback'

Ran 43 tests in 397.347s

FAILED (failures=1, errors=1)
```

bnmnetp:
If I try to run the tests individually, for example `python test_fitb.py` from inside the test folder in fitb I get an error:

```
 from ...unittest_base import module_fixture_maker, RunestoneTestCase
ValueError: attempted relative import beyond top-level package
```

I'm guessing this is somehow related to your `python -m unittest discover` but the error seems right to be `from ..unittest_base` seems like the right path.

bnmnetp:
Bryan,

One more thing...  We need to keep the Runestone Components compatible with both Python 2.7 and 3.x

```
  File "/Users/millbr02/anaconda/envs/web2py/lib/python2.7/unittest/loader.py", line 254, in _find_tests
    module = self._get_module_from_name(name)
  File "/Users/millbr02/anaconda/envs/web2py/lib/python2.7/unittest/loader.py", line 232, in _get_module_from_name
    __import__(name)
  File "runestone/__init__.py", line 10, in <module>
    from .fitb import *
  File "runestone/fitb/__init__.py", line 1, in <module>
    from .fitb import *
  File "runestone/fitb/fitb.py", line 207
    str_num, *list_tol = feedback_field_name_raw.split()
             ^
SyntaxError: invalid syntax
```

bjones1:
To run tests on just one question type, ``python -m unittest discover fitb`` from the root directory of RunestoneComponents. ([Pytest](https://docs.pytest.org/en/latest/) would probably make this even easier, but I haven't checked. We should migrate to that at some point...tests are so much easier to write...)

Re: Python 3, oops! I'm spoiled with my shiny new syntax. I'll fix.

bjones1:
OK, fixed pushed. On Python 2.7, all tests pass (Chrone driver) except one dnd test.

bjones1:
I ran the tests under Python 2.7 and Python 3.6 on upstream/master, and got the same test failures as reported above. Do these fail on Linux as well?

bnmnetp:
When I run this on macOS I get 1 fail and 43 errors.

I'm trying to figure out if its a problem with phantoms or what.  But its hard to debug because the only way I can get the tests to run AT ALL is to run the top level script.  I get import errors when I try to run a single test.

bnmnetp:
@bjones1 

Changing
```
from ...unittest_base import x,y,z
```
to
```
from runestone.unittest_base import x,y,z 
```
Allows me to run the script from the local test directory with the line:

`python -m unittest test_fitb.py`

Once I finally got my old version of chromedriver out of the way all of the fitb tests ran successfully on my mac.

I think we ought to consider the absolute syntax rather than the indirect syntax for the import statements so that we can run the tests individually.  This would be the most common mode for development I would think.

bnmnetp:
@bjones1 don't change the imports.  I've just done it on my local copy of your PR.  I'll merge with master and accept if we get same test results.  Then we'll see what happens on Jenkins with the CI kicks in.

bjones1:
Thanks for the merge. However, I'm certainly in favor of making testing easier -- I'd be very happy to use absolute imports.

bnmnetp:
Good, since I made the change :smile:

In other news, the failure and the error were both problems.  One was where I had mis-placed a close div tag in the dragndrop directive.  A second came from your refactoring where we lost the import of `ActionChains` in the activecode tests.

With those fixes ALL tests are now passing with the Chrome driver.

Hopefully they all pass with the PhantomJS driver on windows too.

In order to help us actually SEE the problems uncovered I've also added a line so that `runestone serve` sends its output to runestone.log file.  -- we could make this a command line option in the future, but this actually seems like decent default behavior.

bjones1:
Whoops -- sorry for injecting a bug. I'm glad that the tests caught this, plus another problem. The PhantomJS drivers on Windows still fail the tests, however. I'm not sure what's going on there, but I'll just use Chrome instead.

bnmnetp:
No worries.  That is what the tests are there for.

It also seems that we can use Chrome in a headless mode. Which would allow us to use it instead of PhantomJS for CI testing.

```
        options = webdriver.ChromeOptions()
        options.add_argument("headless")
        options.add_argument("window-size=1200x800")
        self.driver = webdriver.Chrome(chrome_options=options)  # good for development.
```

bjones1:
Ah, very nice. I trust Chrome a bit more than PhamtomJS, as well...

PR #421, opened 2017-07-17
Fix to Parsons problems without any separators

ericsonga:
I found the error in Jeff's code and added a conditional to protect against it.  When you don't specify any separators it breaks the lines at end of line characters, but that results in one extra empty line.  So I needed to guard against this with a line.length > 0.


bnmnetp:
@aerenchyma Have you been able to make any progress on unit tests for parsons?  I think this is the only one that is missing a selenium test and it would be really great to get that going.

aerenchyma:
@bnmnetp Sorry -- a little bit, but I'm traveling at the moment without a lot of computer/internet access. I'll be back soon and will try to get it going.

PR #425, opened 2017-07-27
Developing 'showeval' Directive

conzty01:


bnmnetp:
I think you should be able to do this with one of the already installed versions of jQuery.  We definitely don't want yet another jQuery to be loaded if we can help it.

PR #428, opened 2017-08-04
Add fix line that got deleted in componentdb

aerenchyma:
I had mentioned on slack that I thought this change was in master, but I just pulled from the RI upstream and lost it again, so perhaps I was pulling it from Paul's and getting confused.

I believe this is the line from componentdb.py that got deleted that's needed for building, so this is a PR to fix that if it's needed.

### DIFF
--- a/runestone/server/componentdb.py
+++ b/runestone/server/componentdb.py
@@ -28,6 +28,7 @@ if all(name in os.environ for name in ['DBHOST', 'DBPASS', 'DBUSER', 'DBNAME']):
     engine = create_engine(dburl)
     meta = MetaData()
     questions = Table('questions', meta, autoload=True, autoload_with=engine)
+    assignments =  Table('assignments', meta, autoload=True, autoload_with=engine)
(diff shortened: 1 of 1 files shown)


aerenchyma:
Thanks @bnmnetp, I have some weird upstream stuff going on that I didn't pick up on until reading this comment. I'll be doing that.

Yes -- though I'm guessing things have changed this summer; I'm getting up to date this week, and didn't realize how far out of date I was on that side of things.

bnmnetp:
There are a number of issues that need to be ironed out with the usageAssignment and possibly assignment directive.  For example, all subchapters are added as questions by the runestone build process now.  Adding them from the usage directive will cause duplicates and potentially other problems.

aerenchyma:
Hmmm, okay. It's working for me now, so I can close this and I think the only remaining thing might be docs.

The assignment directive is definitely broken for builds, I think. usageAssignment I'm not yet sure about problems beyond duplicates. @presnick, thoughts on this?

PR #430, opened 2017-08-09
Misc fixes

bjones1:
Two fixes:

- Make mchoice questions easier to write: use a + for the bullet style of feedback to indicate it's the correct answer. Easier for authors, cleaner code, less messy editing of rest text (used to remove a C or +, which can lead to side effects).
- Have (at least some -- fitb and mchoice) questions tell students if their answers were correct or not when re-visiting a page. Makes it much easier to work on assignments -- they know what's done and what's not yet done.

### DIFF
--- a/runestone/assess/js/mchoice.js
+++ b/runestone/assess/js/mchoice.js
@@ -294,6 +294,11 @@ MultipleChoice.prototype.restoreAnswers = function (data) {
             }
         }
     }
+    if (this.multipleanswers) {
+        this.processMCMASubmission(false);
+    } else {
+        this.processMCMFSubmission(false);
+    }
 };
 
 MultipleChoice.prototype.checkLocalStorage = function () {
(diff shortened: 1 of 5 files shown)


runestone/assess/js/mchoice.js
```diff
@@ -294,6 +294,11 @@ MultipleChoice.prototype.restoreAnswers = function (data) {
             }
         }
     }
+    if (this.multipleanswers) {
```
bnmnetp: What answers were not getting properly restored?  I want to make sure I understand what you are trying to fix here...

bjones1: It's the feedback that isn't being restored -- if students work half a page, then refresh, they won't know if the already-worked problems were correct or not. So, I just invoke the provide_feedback method on the stored answers to show this.

bjones1:
Thanks!

PR #439, opened 2017-08-18
Begin Development of ShowEval v2

conzty01:
Things to do:

- [x] The steps would be embedded into the div either as a variable or as a `<ul>` with `<li>` entries for each step.  
- [ ] The ShowEval constructor would just take the `divid` as the parameter — which is what the others do too.  
- [ ] The trace value would become an attribute of the outter `<div>`  
- [ ] Remove `<script>` tag from HTML doc
- [ ] The constructor would be modified to call the `setNextButton()` on this rather than from the script

### DIFF
--- a/runestone/showeval/showeval.py
+++ b/runestone/showeval/showeval.py
@@ -27,11 +27,11 @@ def setup(app):
     app.add_stylesheet('showEval.css')
 
 CODE = """\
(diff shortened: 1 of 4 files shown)


bnmnetp:
closing as it is outdated and there are other priorities.

PR #446, opened 2017-08-30
Fix: Make Python 2.7 unit tests work.

bjones1:
I'd been testing on Python 3, and wanted to make sure Python 2 works. @bnmnetp, have you tested on Python 2.7? Without this, I get lots of failures on Windows.


bnmnetp:
runestone.academy runs live on 2.7

bjones1:
Strange. I don't understand why this is needed, but my tests on Python 2.7 error without it.

PR #447, opened 2017-08-31
Do not include the "correct/total" results of dragndrop in logged answers

shadaj:
This line appears to be mistakenly added in https://github.com/RunestoneInteractive/RunestoneComponents/commit/0561b8372365e9b6e5d1c32b9f0fc407eed8ae96, which broke the dragndrop directive by causing an exception when loading saved answers and preventing dragndrop answers from being autograded.

Potential fix for https://github.com/RunestoneInteractive/RunestoneServer/issues/1025

### DIFF
--- a/runestone/dragndrop/js/dragndrop.js
+++ b/runestone/dragndrop/js/dragndrop.js
@@ -348,6 +348,7 @@ DragNDrop.prototype.dragEval = function (logFlag) {
             "minHeight": this.minheight,
             "div_id": this.divid,
             "correct": this.correct,
+            "correctNum": this.correctNum,
             "dragNum": this.dragNum
         });
     }
--- a/runestone/dragndrop/js/dragndrop.js
+++ b/runestone/dragndrop/js/dragndrop.js
@@ -341,7 +341,6 @@ DragNDrop.prototype.dragEval = function (logFlag) {
     this.renderFeedback();
     if (logFlag) {  // Sometimes we don't want to log the answers--for example, on re-load of a timed exam
         let answer = this.pregnantIndexArray.join(";");
-        answer = answer + ":" + this.correctNum + "/" + this.dragNum ;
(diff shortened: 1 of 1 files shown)


bnmnetp:
And thanks for the PR!  Very much appreciated.

shadaj:
@bnmnetp that makes sense! I'll work on placing the percentage under a separate key.

shadaj:
@bnmnetp ready for another review!

runestone/dragndrop/js/dragndrop.js
```diff
@@ -341,7 +341,15 @@ DragNDrop.prototype.dragEval = function (logFlag) {
...
         let answer = this.pregnantIndexArray.join(";");
-        this.logBookEvent({"event": "dragNdrop", "act": answer, "answer":answer, "minHeight": this.minheight, "div_id": this.divid, "correct": this.correct});
+        this.logBookEvent({
+            "event": "dragNdrop",
+            "act": answer,
+            "answer": answer,
+            "minHeight": this.minheight,
+            "div_id": this.divid,
+            "correct": this.correct,
+            "dragNum": this.dragNum
```
bnmnetp: I'm not sure how dragNum helps us, it really only tells how many things were available to drag.  What I really want back at the server is that number divided by the number that were correct -- this.correctNum

Brad

shadaj: The idea was that with dragNum we have all the information we need to calculate the percentage on the server-side, instead of having the client preprocess the data.

bnmnetp: That would be fine, but I don't think we do have enough information.  Correct is a boolean. and answer only tells us the matchings.  The server doesn't know which matches were correct.

shadaj: Oops, missed that correct and correctNum are different values!

PR #448, opened 2017-09-03
Pr video https option

dirkcgrunwald:
Added option to have video be either https or http with default to http. This is needed when viewing a runestone document through an iframe on a site using https -- without it, the enclosed videos will not show (at least one chrome). Both Youtube and Vimeo support https so making this the default shouldn't change behavior and is more compatible with "encrypt everywhere".

### DIFF
--- a/runestone/video/video.py
+++ b/runestone/video/video.py
@@ -171,7 +171,7 @@ class IframeVideo(Directive):
         'height': directives.nonnegative_int,
         'width': directives.nonnegative_int,
         'align': align,
-        'align': httpOption,
+        'http': httpOption,
     }
     default_width = 500
     default_height = 281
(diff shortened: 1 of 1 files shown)


bnmnetp:
Thanks Dirk,

The video directives all seem to work for me on both Chrome and Safari when I’m on https://runestone.academy

Where are we showing videos in an iframe?  Or is this on a custom book you are working on?

I think the option is a great idea, I just want to make sure I haven’t missed something on the academy site.

dirkcgrunwald:
This is for a custom book on of our instructors is working on. The content is served from an NGINX web server which forces https ( https://runestone.stage.csel.io/thinkcspy1200/ ). Students access this from our LMS (moodle) and I think that's where the iframe comes up.

PR #449, opened 2017-09-06
Parsons: block numbering, help and indent changes

CABowers:
Blocks/paired distractors are now numbered. Changes to the order in which HelpMe provides help. And removed indentation in answer area if the problem has no indentation.


ericsonga:
I would like to have these changes integrated into Runestone Components before I do my next experiment.

bnmnetp:
Barb, when will the next experiment run?

ericsonga:
I hope to run it by mid to late Oct.  But, there are still some more changes I need to make to the Parsons problems before then, so I would like these changes done so that I can add the others as well.

ericsonga:
My first experiment is Oct 22nd.  Can we please integrate this pull request before Oct 15th, so that I can create my experiment materials and test them?

bnmnetp:
Should we make the numbers optional / not there when there are no paired distractors and no help?  I think they are confusing.  Or maybe we could style them differently so they don't stand out so much.  I think a kind of gray outlined font would not be so forward.

bnmnetp:
Also, are the unittests passing for you??  The parson's unittests are failing on my mac.

CABowers:
They were passing on my Windows laptop, which case(s) are failing?

CABowers:
This could have been the issue, my laptop is slow so it is possible that animations had time to finish before the next step was run. Regardless, this change should make it more robust.

ericsonga:
There is also a problem with another class being used sometimes for the label.  It is class lit.   This makes it a dark blue color.  Cole can you figure out what is going on with that?

I agree that the color would be better if it was a gray.  Cole change it to 

.label {
	font-family: "Courier";
	font-size: 8px;
	color: #a6a6a6;
	position: absolute;
	right: 5px;
}

bnmnetp:
OK, unit tests are passing but styling is not updated yet.

I'm still concerned that the numbers will cause more confusion than help for the students, especially for simple problems without paired distractors.

CABowers:
I have updated the style based off of a meeting I had with Barb.  I can work on an option to turn the numbering on and off.

runestone/parsons/css/parsons.css
```diff
@@ -252,10 +252,28 @@
...
+.block-label {
+	font-family: "Courier";
+	font-size: 10px;
+	color: #b6a6a6;
+	padding: .2em 0.0em .3em .6em;
+	vertical-align: top;
+	white-space: pre;
+}
+
+.labels {
```
bnmnetp: So you removed .label and changed it to .labels

But the element in the parsons problem still has a class of .label 

.label is not a good choice for this name as it is already used by bootstrap.  So you should pick a different name and update the generated parsons element to use the new class name as well.

CABowers: I did realize that .label was reserved so I switched it to .block-label. I use .labels to style the the div that holds all the labels for a given block. I don't see where I am still adding .label to an element. Could you show me?

bnmnetp:
OK, my misunderstanding.  But It doesn't look like the styling has changed, and I've cleared my cache to be sure I"m getting the right css file.  @ericsonga I know you want this accepted soon, but can we get agreement on when/where the numbers should be used?

CABowers:
I am unsure why the style isn't changing for you. Did you update Parsons.js as well as the css?

ericsonga:
Cole will finish the rest of the changes he needs to do today.  I will be running my experiment on Sunday from 3pm to 6pm.  I would really like his changes integrated by Wednesday.  I will have materials for you to upload as new books (just for the experiment).  

I think the numbers are useful for people talking about the blocks, but for now we will only put them on the Parsons problems with paired distractors.  Cole will also add the numbers as an option but the default will be no numbers.

PR #481, opened 2017-10-15
Fix: Detect and report duplicate divids in Runestone directives.

bjones1:
Also, DRY code -- consolidate self.options['divid'] = self.arguments[0].


bjones1:
Oops, this is wrong. Closing.

PR #486, opened 2017-10-21
Simplified and clarified some instructions for running tests

DaveParillo:


DaveParillo:
Good catch.  I'll close this and resubmit

PR #487, opened 2017-10-21
Readme tests typo

DaveParillo:
Simplified and clarified some instructions for running tests.

The old instructions do not work.


bnmnetp:
Thanks for the updates!

PR #488, opened 2017-10-21
added fitb unit tests for casei option and numeric ranges

DaveParillo:
I wasn't sure I completely understood the regex handling in the answer fields, so no tests for that.

But does have some tests for the `:casei:` option, implied blanks, and numeric ranges.

### DIFF
--- a/runestone/fitb/test/_sources/index.rst
+++ b/runestone/fitb/test/_sources/index.rst
@@ -65,3 +65,20 @@ Regex testing
        :2 1: Close.... (The second number is a tolerance, so this matches 1 or 3.)
        :x: Nope. (As earlier, this matches anything.)
 
+.. fillintheblank:: regexescapes1
+   :casei:
+
+   Windows system files are stored in: |blank|. 
+
+   -   :C\:\\Windows\\system: Correct.
+       :program files: Third party applications are stored here, not system files.
+       :x: Try again.
+
+.. fillintheblank:: regexescapes2
+   :casei:
+
+   Python lists are declared using: |blank|. 
+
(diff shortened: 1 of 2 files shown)


bnmnetp:
Thanks, its great to have more coverage.

What do we need to explain better with respect to the regex matching?

DaveParillo:
There are many different syntax variations for regex.  But even presuming that the underlying parser is python, it's not clear what get's done with the text in the field `: x :`. The source code hints that reular expressions are possible, but other than `:.*:` or `:^foo$:`, wasn't sure what was possible. 

Some trial and error quickly determines no flags can be used before or after the `:text:`, but what text is allowed? If the answer is 'simple case insensitive string matches', then great.
If the answer is some thing else, then some power users may want to know how to do that.

Is the match function always invoked?  Since `:casei:` is a directive option, it implies that it is the only regex flag that can be set.  LOCALE and UNICODE might be nice additions.

If you defined a more complex regular expression, say with a grouping structure, how would you inform runestone what the expected value for group 1 should be?  This seems hard.
And perhaps trying to do more than one little fill in the blank field should be trying to accomplish.

DaveParillo:
After giving this a bit more thought, I realized two baked in assumptions about fitb:

1. The first answer is always considered correct & the remaining sub items within a list are always incorrect.  The help should explicitly state this.
1. If an answer field _looks like_ a regex, it is parsed as a regex, meaning some answers such as:

    - `C:\windows`
    - `myarray[x][y]`

will require some escaping. But I haven't tested this.

I think I have now sufficiently over analyzed this one!

PR #490, opened 2017-10-24
Change the style of tabpanel to fit the width of container

lisheng0316:


lisheng0316:
Thanks for the review, here is what has changed

before1
after1
before2
after2

bnmnetp:
This seems to be redundant with the work we already did to implement the ``full-width`` class.???

PR #501, opened 2017-10-31
Audio Tour Revamp

PC98:
1. Changed the audio tour UI - instead of a separate screen popping up, the audio tour now launches in the same screen, right below the "Audio Tour" button.

2. Because the audio tour doesn't require a separate screen to pop up now, the user can have audio tours from multiple coding examples open simultaneously.

3. Earlier, images were being used as place-holders for media control buttons (that is, the "play button", "next-audio button" etc. were images); I have replaced them with actual HTML button tags.

4. A bunch of bug fixes, for example, earlier audio would play when the user clicked on the "next audio" button even though "pause audio" was selected.

5. Corrected the URL path for audio files.

I have tested the new audio tour on Chrome, Safari and Firefox.


bnmnetp:
It would be great if there was some way we could automate the testing of the audio tours.  Any ideas?

bnmnetp:
@ericsonga Can you please take a look at this.  It seems that this PR also contains code for resetting timed exams. Does this replace what you did?  Is it the same?  This is why we need to have 1 feature per PR.

PC98:
Hi!
I am not sure if there is a way to automate testing of the audio tours.

This PR contains code related to timed exams because of a mistake on my part - I pulled from CSL lab's fork before pushing my changes. If you would prefer it, I can close this PR and open a new one with just the audio tour changes.

bnmnetp:
It would be great if you can clean this up so that it contains only the code related to the audio tours.  then I'll accept it.

ericsonga:
Yes, let's only do the audio tour in the pull request.

PC98:
Hi. Sorry I have been inactive for the past couple of weeks. I am not sure what the best way to do this is, but I think I will create a new fork, merge only my changes and then make a pull request.

PR #505, opened 2017-11-22
Supplementary files for Java

CABowers:
I closed my other PR related to this and brought everything together to make it neater. I made changes to the server to allow more jobe api calls to go through the proxy, those need to be added for these changes to work.


bnmnetp:
please **add** a test to the activecode tests to demonstrate that this works and will allow us to catch regressions in the future.

bnmnetp:
We should modify  the LiveCode.prototype.init method so that `this.JOBE_SERVER = eBookConfig.host`  uses `eBookConfig.jobehost` and also get this.resource from `eBookConfig.proxyuri`    These could be set up as two new variables in the pavement.py file.  The main template (layout.html) would need to be modified as well to add these to eBookConfig.

CABowers:
I am unsure what `eBookConfig.jobehost` should be set to. Also should I be making those changes to the `pavement.py` and `layout.html` in the `activecode/test` and to `pavement.tmpl` and `layout.html` in `common/project_template`?
Another issue is if the tests are running from `127.0.0.1:8081` and the server running from `127.0.0.1:8000`, wouldn't this be a cross origin request for the proxy functions? A solution could be on the server side to add `127.0.0.1:8081` as the only address allowed to make cross origin requests just for testing purposes but the flaw with this is addresses can be spoofed from a security perspective. What do you purpose?

bnmnetp:
Before we created the proxy we could access the jobe host directly.  In part because we have an API key that we use to authenticate ourselves.  The proxy is needed because the jobe host runs on http not https.  So the CORS issues are dealt with.

Here is the information you need for accessing the jobe server directly.

```
JOBE_SERVER = 'http://jobe2.cosc.canterbury.ac.nz';
resource = '/jobe/index.php/restapi/runs/';
```

This change will need to be made in all of the places you mention.

CABowers:
I tried to resolve the conflicts via the interface on the website and I resolved them in my most recent commit. However, it still shows the code as conflicting. When I view this PR from a different view it seems that it believes there are no conflicts as seen in the below. 
image
I will make a duplicate branch and make a PR from that branch to see if it re-evaluates the conflict.

bnmnetp:
Some cleanup is in order.    This should not require 200 files modified.  It seems most of them are under activecode/test/ . An extra static directory has sprouted.  Also it looks like there are massive changes to activecode.js . Can you clean this up so that only the required changes are part of the PR?

Messing with the runProg function and the history scrubber makes me *extremely nervous* as it took a long long time to get the bugs worked out of the initial implementation and if it breaks the production version I'm going to have thousands of unhappy students.

PR #506, opened 2017-11-22
Made the numbered Parsons blocks a directive option defaulted to off

CABowers:


bnmnetp:
We are missing the conf.py in the test folder.   Also, the index.rst file can have tests in it for both numbering on and numbering off.

CABowers:
Thanks for doing that. I'll definitely get those tests done, I had to take a break to study for finals but I'll be done soon and can spend time again getting the fixes finished up.

Issue #510, opened 2017-12-13
Activecode download 


kooi:
I would like to I've students the ability to download the code from an activecode block. This gives students the ability to backup / export their work to other systems easily.

bnmnetp:
Thats an interesting idea.  

Are you thinking that this would work for a single activecode?  There would be a button that would prompt them to enter a name and then they would get a name.py file in their download folder filled in with the code from that activecode window?

I always just have my students copy and paste.  What are the advantages of having a download link over the copy/paste approach?

kooi:
I've actually managed to get this working; I extended the activecode definition with an :enabledownload: flag so you can turn this off and on. If this is a useful feature I'll do a pull request.

I considered the copy-paste method which I thought would be no problem, but I've been working with kids (age 11-12) with a very varied skillset; most don't really seem to have trouble grasping the basic programming concepts but basic computer skills are sorely lacking in many (the concept of copying and pasting is alien?!) and this seemed an easy solution to hand in files via moodle without needing LTI-integration (which personally I think I would prefer).

bnmnetp:
Indeed LTI integration is on the roadmap for a couple of years.  I think it (LTI) may have gotten to the point where we can really make it a high priority this summer.

In the meantime, I would be happy to accept a PR for this.

Thanks!

PR #521, opened 2017-12-27
Activecode download button

kooi:
Added the ":enabledownload:" flag to the activecode directive to enable Blob download of activecode data.
(issue #510 )

Currently automatically names the file as "<activecode-id>_<reversedate>.py" 
`var fileName = fnb + '_' + d.toJSON().substring(0,10).split('-').join('')+'.py';`

(only activecode.py and activecode.js should be changed; setup.py is only edited for checking purposes)


bnmnetp:
Thanks for this.

It looks like you have a couple of conflicts to clean up.  And some extraneous console.log messages to get rid of.

On a different level, this seems like the kind of thing where a global flag such as ``eBookConfig.downloadsEnabled`` would be useful.  If an option was specified on the directive explicitly then that should take precedence, but if no options was there either way then letting the instructor enable/disable this for everything would be really nice.

kooi:
A global flag seems like a good idea, I'll get on that.

kooi:
I think I enabled the global flag correctly, though I'm not entirely sure but it does seem to work for me.
I also read the now implement the correct default file extension based on the Activecode language setting.

bnmnetp:
Well done, thank you.

I will accept this now.  It may take a couple of days to update thinkcspy and then update the website so that folks could enable this for their courses.

PR #576, opened 2018-05-01
Internationalization

vasiljevic:
Implementation of general internationalization features for the front end code (JavaScript and HTML) based on Wikimedia'a [jQuery.i18n](https://github.com/wikimedia/jquery.i18n) and internationalization of of the ActionScript component (with partial Serbian translation).

More detailed description of general internationalization features with internationalization guidelines you can find in I18N.md (contained in this pull request).


bnmnetp:
I'm testing this out on my dev machine today.  It all looks good, except when I build the font in the activecode windows is tiny.  Any idea why that might be happening?

vasiljevic:
@bnmnetp I have reproduced the issue, but it seams that the same issue exists in the main branch i.e.  looks like the issue is not related with the pull request.  

BTW, CodeMirror just skip to transform the Python source code into the highlighted form. I have tried to dig inside but I'm not familiar with the CodeMirror.

vasiljevic:
@bnmnetp I have found what is the problem with CodeMirror: the WebglDemo directive also adds CodeMirror JavaScript files, but without python.js. 
After codemirror.js is loaded second time, it resets the list of loaded mods, and the Python mod stays unloaded.
I have made a quick fix commit in this pull request: WebglDemo adds python.js too. Python mod in CodeMirror editors works fine now, but CodeMirror scripts are still unnecessarily duplicated in the HTML.

bnmnetp:
Great find,  Thanks for saving me the trouble!

PR #595, opened 2018-06-06
Fix test failures

bjones1:
On Windows, with Sphinx 1.7.4, these fixes are required to get the unit test suite to pass.


bnmnetp:
@presnick Please take a quick look at this one as it effects changes from you we just incorporated to account for chapter or subchapter names that have /'s in them.

@bjones1 we need to make sure this works for Paul's use case as well.

bjones1:
Sounds good. However, if everything is working under Docker/Linux, this shouldn't matter, since `os.sep == '/'` on those platforms.

PR #605, opened 2018-06-16
Fix of the accessibility workaround

vasiljevic:
This is a longer-term fix of the workaround from the commit  c65db20982bb7497bae056a80afd716c05d8b107  (Added Acessibility.css files) 

In the commit message @DomKynkor has noted: "I couldn't get accessibility.css to generate an HTML reference when the files were built, so I included a work around of importing accessibility.css into codemirror.css"

This pull request encapsulates accessibility stylesheets in an extension and also introduces a config value to select what stylesheet file to add.


bnmnetp:
Thanks, I really appreciate your PR's and work on the internationalization.

Could you give me an overview of what this is doing?  Does it impact the accessibility work that was don on parsons problems?

vasiljevic:
This is just a better place for the same Acessibility*.css files. A kind of refactoring, no impact on the accessibility itself. In default configuration it should provide equivalent output to the existing one. Additionally you can select by a config value one of the provided Acessibility*.css files (currently, you have to modify codemirror.css to do it). Also, you can select not to include any  of the provided Acessibility*.css in your book project if those stylesheets are not compatible with a theme you use.

PR #606, opened 2018-06-16
Avoid repeating same database error in activecode.py

vasiljevic:
Some authors use the `runestone` command just to preview their contributions and don't need to configure databases on their computers. In this scenario, the `activecode` directive emits the same error for each occurrence of the directive:
```
Unable to save to source_code table in activecode.py. Possible problems:
  1. dburl or course_id are not set in conf.py for your book
  2. unable to connect to the database using dburl

This should only affect the grading interface. Everything else should be fine.
```
With this small PR the `activecode` directive will report that error only on the first occurrence.

### DIFF
--- a/runestone/activecode/activecode.py
+++ b/runestone/activecode/activecode.py
@@ -361,11 +361,13 @@ config values (conf.py):
                 available_files = self.options.get('available_files', "")
             ))
         else:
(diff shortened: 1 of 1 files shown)


bnmnetp:
I should have thought of this one a long time ago!

Thanks.

PR #610, opened 2018-06-18
Bug fix for CodeLens; improvements for presenter view

presnick:


runestone/activecode/js/activecode.js
```diff
@@ -611,7 +611,7 @@ ActiveCode.prototype.showCodelens = function () {
 
     var cl = this.codelens.firstChild;
     if (cl) {
-        div.removeChild(cl)
+        this.codelens.removeChild(cl)
```
presnick: This is just fixing a bug. It wasn't removing it, so it wasn't doing a new call to create it with the new code.

runestone/activecode/js/activecode.js
```diff
@@ -437,7 +437,7 @@ ActiveCode.prototype.createOutput = function () {
 
 
     var lensDiv = document.createElement("div");
-    $(lensDiv).addClass("col-md-6");
+    $(lensDiv).addClass("col-md-12");
```
presnick: This was needed for presenter view and we don't think it hurts anything in regular textbook mode.

PR #615, opened 2018-06-20
Misc fixes

bjones1:
This is a grab bag of fixes:
- Remove a `.gitignore` entry that hides `conf.py` files used for testing.
- Include tests by supplying a missing `conf.py` and `__init__.py`.
- Fix broken tests.
- Fix Unix-only code in tests.
- Correctly include Parson's problem JS libraries.
- Remove an old and broken JS path in the template.
- And more...

However, both before and after these tests, I still have test failures in activecode and Parson's problems. Someone else, please fix these! Don't commit files that fail tests!


runestone/parsons/parsons.py
```diff
@@ -25,9 +25,9 @@ def setup(app):
     app.add_directive('parsonsprob', ParsonsProblem)
     app.add_node(ParsonsNode, html=(visit_parsons_node, depart_parsons_node))
     app.add_stylesheet('parsons.css')
-    app.add_stylesheet('lib/prettify.css')
-    app.add_javascript('lib/prettify.js')
-    app.add_javascript('lib/hammer.min.js')
+    app.add_stylesheet('js_lib/prettify.css')
+    app.add_javascript('js_lib/prettify.js')
+    app.add_javascript('js_lib/hammer.min.js')
```
bnmnetp: What does this fix?

runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html
```diff
@@ -11,16 +11,11 @@
     '_static'+'/processing-1.4.1.min.js',
     '_static'+'/jquery.hotkey.js',
     '_static'+'/jquery-migrate-1.2.1.min.js',
-    '_static'+'/presenter_mode.js',    
+    '_static'+'/presenter_mode.js',
   ]
 %}
 
 {% set css_files = ['_static/bootstrap-3.0.0/css/bootstrap.min.css', '_static/presenter_mode.css'] + css_files %}
-{% if theme_bootswatch_theme %}
```
bnmnetp: This does not break anything??!!

runestone/parsons/parsons.py
bjones1: Oddly, when I look at the files in `_static/` produced by Sphinx/runestone build, there is no `lib/` subdirectory and the .js files that should be in `lib/` aren't present. The rename means that the `js_lib/` subdirectory and all the expected files are present. This is all when using Sphinx 1.7.4. It's so odd that using `lib/` doesn't work, but it just doesn't. Can you verify this behavior on Linux/Mac?

runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html
bjones1: Just the opposite, it fixes something. Before this patch, I saw the line:

`    <link rel="stylesheet" href="http://netdna.bootstrapcdn.com/bootswatch/2.3.1/""/bootstrap.min.css" type="text/css" />
`

in all the output HTML files. The deletion in this PR removes these (broken) lines from the output.

runestone/parsons/parsons.py
bjones1: Oops, I'm (partly) wrong on this. On a branch I'll (eventually) merge, this bug appears and this PR fixes it; however, it doesn't appear when not using my branch. I'll do more investigation.

bjones1: OK, I figured this out. I had a `lib/` directory containing library files to exclude from the build, but this also excluded all JS in the `lib/` directory of the Parson's problems. Ouch.

To avoid namespace/directory naming problems like this, I'd suggest keeping this change (use the name `js_lib/`) to make it a more distinct name. @bnmnetp, what's your preference?

bnmnetp: Is there any reason that those files are in a lib folder?  Why can't they all just live directly in the js folder?

bjones1: Good point. I'm not the author of the Parson's problems, but FWIW: the `lib/` subdirectory includes files not written by Runestone people, but copied from other projects. Hence, the idea of putting these files in a place to indicate they're a dependency makes sense...but perhaps using another name.

runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html
liffiton: I think that might be because of an issue in `theme.conf`, not a problem here.  I ran into an issue where if the `bootswatch_theme` variable is set to `""` (as is done in the included `theme.conf` and as [the comment above](https://github.com/RunestoneInteractive/RunestoneComponents/blob/fa19c4bdff33649145d6683c507338c56f7499b7/runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/theme.conf#L42) suggests is the default), then it includes the `""` you're seeing.  I believe the line in `theme.conf` should be `bootswatch_theme = `, without any quotes, to use no theme.  Just based on my recollection, anyway.

liffiton: Well, I should probably add to that:  The template pulls in Bootstrap locally from the static/ folder somewhere else, hence why it doesn't cause problems when the CDN bootswatch CSS fails to load.  If you correct the `bootswatch_theme` setting, then pages will pull in Bootstrap twice (and I think with two different versions).  So it's probably best to simplify by either dropping the bootswatch theme loading altogether and just use the local Bootstrap, or to move entirely to CDN bootstrap and drop the locally-hosted version.

bnmnetp:
@bjones1 All unit tests run on master for me on macos and on Linux via Jenkins.

I'm going to set up a Travis-CI task so that they are run automatically when a PR is created or something is committed.  I'll try to follow my own workflow a little more diligently about making PR's myself rather than committing directly to master.

runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html
bjones1: @liffiton, thanks for the feedback. IMHO, if Runestone should be capable of working locally (without a server), then Bootstrap should be stored locally. So, I'll amend this PR by removing the incorrect settings in `theme.conf`, as well as the removal already done in this PR.

bjones1:
@bnmnetp, thanks for testing on your side. I re-ran the tests and still see failures:

```
FAIL: test_datafile (runestone.activecode.test.test_activecode.ActiveCodeTests)
  File "E:\Runestone\RunestoneComponents\runestone\activecode\test\test_activecode.py", line 78, in test_datafile
    self.assertLess(count, 20)
AssertionError: 20 not less than 20
```

I'm running on Python 3.6.1, Sphinx 1.7.4.

Running this manually in the browser, I see an error in the browser's JavaScript console: `Failed to load http://jobe2.cosc.canterbury.ac.nz/runestone/proxy/jobeCheckFile/runestone1278007241665d682b16344a5017be5e: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'null' is therefore not allowed access.`

Perhaps this is a newer error based on updated access control policies?

bjones1:
I get the same failures with Python 2.7.13. All this was tested using Chrome 67.

bnmnetp:
@bjones1 Have you synced with master?  I did push a small change 072b0127abdd58b4fbf17b3fb3173755c249a7a3 to fix this.

bjones1:
@bnmnetp, thanks -- that fixed it; all tests now pass on Python 2.7 and 3.6 for me. When running the tests using `pytest` instead of `unittest`, I see one failure, but I haven't investigated this. I went ahead and rebased to include your fixes.

bjones1:
I figured out the Parsons's testing failures: the `parsons/` directory is missing an `__init__.py`, so its tests are not being run. It looks like these questions are broken -- the question number get embedded with the draggable boxes. @bnmnetp, would you like to me inclucde an `__init__.py` in this PR? If so, it will reveal the failing tests.

image

bnmnetp:
@bjones1 lets leave them disabled, but could you file a PR on this.  I think there are two problems that this demonstrates and I would like @ericsonga to investigate

1.  adding a bit of text  before the first `-----` removes the numbers from the problem but...
2. the problems then become double numbered.

I have quick fixes for this that I can do as a separate PR.

bjones1:
@bnmnetp, will do in a separate PR!

bjones1:
@bnmnetp, since this was merged, do you want a PR/followup on [your comment](https://github.com/RunestoneInteractive/RunestoneComponents/pull/615#issuecomment-403465382)? The missing `__init__.py` for Parsons's tests was added when you merged this PR.

bnmnetp:
Done, See #632

bjones1:
OK, thanks -- just didn't want to drop the ball.

PR #617, opened 2018-06-30
Several updates/simplifications to the datafile directive

liffiton:
- Add a label indicating the filename (for reference in activecode)
 - Make styling of editable datafile consistent with other `<pre>`s
 - Remove need for javascript and the "flash" on load for hidden datafiles
 - Cleanup python code

Example:

image

I should say up front that I'm not sure this follows RSE-0001, exactly.  It removes the use of any javascript for things that can be directly set in the HTML, which avoids things like content flashing on the page briefly before the javascript can hide it and inconsistent styling between editable and non-editable datafiles.  And so it is *not* backwards compatible for anyone who is solely using the javascript (and not the Sphinx/rst directives), because the intent is that the tag for the datafile can just be written in the HTML directly with no need for javascript to touch it.  I'm not sure if that kind of backwards compatibility is a requirement or not.  It is backwards compatible as far as I can tell when using the directive.

If it doesn't fit how you want to do things, no problem, and we can just close the pull request.  Parts could be manually cherry-picked without too much trouble:  You could adapt the label and related CSS independently of the rest, and the existing javascript could be changed to leave everything as a `<pre>` for consistent styling.


bnmnetp:
Thanks for this @liffiton !  Its always great to get something from a new contributor.

I like the new feature of the label that spells out the filename.

And I always appreciate simplification where we can, however there are a few concerns I have on this, due to the increasingly complicated way in which many of the directives interact with the server.

1.  All the generated directives are now stored in the database both in rst form and the generated HTML.  We need to make sure this is preserved.
2. datafiles in particular need to go into the database to be sure that they are accessible from assignments and from the grading page, timed exams etc.  
3. It also makes this directive different and inconsistent in a way which I'm pretty sure would break the timed exams as they rely on the `component_factory` to do the proper hiding and revealing inside an exam.

Maybe your simplification addresses all of that, but I didn't see it in my first look.

liffiton:
The modifications don't touch the database insert/update, as far as I can tell.  It's still there in `__init__.py`.  I haven't seen anything that would save the generated HTML in the database.  Is there some other part of the extension that interacts with the database?  If so, I can make sure it's unaffected as well.

I can't see how the `component_factory` interacts with the datafile extension; `datafile.js` didn't have any mention of `component_factory`, unlike the other extensions.  But I can't find any documentation or definition for `component_factory` in this repository or in RunestoneServer, so I seem to be missing something.  If you have a pointer about that, I can look into it.

bnmnetp:
Thanks again.  I cherry picked the nice tab.  But I"m going to keep the current implementation as I think that better fits within the direction I would like to take this.  Specifically I've been exploring how to make the components or at least some of them work within the Jupyter notebook environment.

PR #642, opened 2018-07-25
Fixes and improvements for numbered problems

bjones1:
Fix: Restart numbering at 1 on every page.
Fix: Directives that begin with a directive now render correctly.
Add: Visually-checked example of a directive which begins with a directive.

### DIFF
--- a/runestone/common/runestonedirective.py
+++ b/runestone/common/runestonedirective.py
@@ -16,14 +16,15 @@
 __author__ = 'bmiller'
 
 
-import os
 from collections import defaultdict
 from docutils import nodes
 from docutils.parsers.rst import directives
 from docutils.parsers.rst import Directive
 from docutils.utils import get_source_line
+from docutils.statemachine import ViewList
 
 UNNUMBERED_DIRECTIVES = ['activecode', 'reveal', 'video', 'youtube', 'vimeo', 'codelens', 'showeval', 'poll', 'tabbed', 'tab', 'timed', 'disqus']
+
 
 # Provide a class which all Runestone nodes will inherit from.
 class RunestoneNode(nodes.Node):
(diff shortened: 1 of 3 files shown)


bnmnetp:
Are we sure this is what we want to do?  I could see a use case where people would want to number exercises sequentially for an entire Chapter rather than starting at 1 on every subchapter.  I think this depends a little on whether you prefer to write with long pages or lots of short pages.  But I could see it getting confusing if there were multiple list-q-1 in the lists chapter.

bjones1:
The current system is broken; this at least gives consistent numbering. In particular, consider this case:

```
Chapter 1
  Subchapter A
  Subchapter B
  Subchapter C
```

After a `runestone build --all`, everything is OK. Then, an edit of subchapter B will produce renumbered problems in that subchapter only -- and the renumbering will either start at 1 (if the previous numbering wasn't saved in `env`) or the next number after the last problem on subchapter C (if not -- I don't remember which). Adding or removing problems to subchapter B should cause a renumbering of subchapter C, but it doesn't. Fixing these problems seem fairly challenging to me, so I opted for the simpler fix presented here.

runestone/common/runestonedirective.py
```diff
@@ -94,6 +98,10 @@ def _purge_runestone_data(app, env, docname):
         runestone_data.id_to_page.pop(id_)
     runestone_data.page_to_id.pop(docname)
 
+    # Reset the problem numbering for each page.
+    env.assesscounter = 0
```
bnmnetp: if the .. qnum directive explicitly sets the starting value for this chapter will this override it??

bnmnetp:
OK, good point.  We actually addressed that in fopp by setting the prefix for each subchapter.

I just want to make sure this doesn't override an explicitly set starting number for a chapter.

runestone/common/runestonedirective.py
bjones1: No, this occurs before any directives on the page are processed. It's used to purge any stale data for the page (hence the name).

bjones1:
Sounds like a good chance to test the `qnum` directive. I can add this to the unit test (although checking it will be harder...)

bjones1:
I've just pushed updates which includes tests of `qnum`, plus does better checking of the directive-in-directive case. This should be ready to go!

bnmnetp:
Thanks!

PR #644, opened 2018-07-31
adjust error line numbers for included code

bnmnetp:
This updates the line number that appears in an error message to account for included code.

### DIFF
--- a/runestone/activecode/js/activecode.js
+++ b/runestone/activecode/js/activecode.js
@@ -694,12 +694,27 @@ ActiveCode.prototype.toggleEditorVisibility = function () {
 
 ActiveCode.prototype.addErrorMessage = function (err) {
     //logRunEvent({'div_id': this.divid, 'code': this.prog, 'errinfo': err.toString()}); // Log the run event
+    let errorOutside = false;
+    if (err.traceback.length >= 1) {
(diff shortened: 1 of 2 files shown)


runestone/activecode/js/activecode.js
```diff
@@ -693,13 +693,31 @@ ActiveCode.prototype.toggleEditorVisibility = function () {
...
-    //logRunEvent({'div_id': this.divid, 'code': this.prog, 'errinfo': err.toString()}); // Log the run event
+    // Add the error message
+    // But, adjust the line numbers.  If the line number is <= pretextLines then it is in included code
+    // if it is greater than the number of included lines but less than the pretext + current editor then it is in the student code.
+    // adjust the line number we display by eliminating the pre-included code.
+    let errorOutside = false;
+    if (err.traceback.length >= 1) {
+        errorLine = err.traceback[0].lineno;
+    }
+    if (errorLine <= this.pretextLines || errorLine > (this.progLines + this.pretextLines)) {
```
bjones1: It seems to me that `errorLine` will not always be defined here -- shouldn't this code go inside the above `if` statement?

bnmnetp: You are correct.  Thanks.

PR #646, opened 2018-08-02
Feedback colors

bnmnetp:
Using red and green for incorrect and correct feedback is problematic for color blind folks.

I'm proposing the following:

eczzmuddvsg

### DIFF
--- a/runestone/assess/js/mchoice.js
+++ b/runestone/assess/js/mchoice.js
@@ -428,10 +428,10 @@ MultipleChoice.prototype.renderMCMAFeedBack = function () {
     var feedbackText = this.feedbackString;
 
     if (this.correct) {
-        $(this.feedBackDiv).html('Correct.<ol type="A">' + feedbackText + "</ul>");
-        $(this.feedBackDiv).attr("class", "alert alert-success");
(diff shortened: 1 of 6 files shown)


runestone/assess/js/mchoice.js
```diff
@@ -428,10 +428,10 @@ MultipleChoice.prototype.renderMCMAFeedBack = function () {
     var feedbackText = this.feedbackString;
 
     if (this.correct) {
-        $(this.feedBackDiv).html('Correct.<ol type="A">' + feedbackText + "</ul>");
-        $(this.feedBackDiv).attr("class", "alert alert-success");
+        $(this.feedBackDiv).html('✔️ <ol type="A">' + feedbackText + "</ul>");
+        $(this.feedBackDiv).attr("class", "alert alert-info");
     } else {
-        $(this.feedBackDiv).html("Incorrect.    " + "You gave " + numGiven +
+        $(this.feedBackDiv).html("Not quite.    " + "You gave " + numGiven +
```
bjones1: A minor nitpick -- why not have an X here to match the checkmark above? That seems more consistent to me. Other than that, I'm fine with this.

PR #665, opened 2018-09-05
Allow multiple youtube videos on one page

ascholerChemeketa:
Modified the class Youtube to directly build each video as an iframe rather than using the IFrame Player API. (https://developers.google.com/youtube/player_parameters). The existing IFramePlayer implementation only rendered the last video on any page.

### DIFF
--- a/runestone/video/video.py
+++ b/runestone/video/video.py
@@ -198,11 +198,7 @@ class IframeVideo(RunestoneIdDirective):
         raw_node.source, raw_node.line = self.state_machine.get_source_and_line(self.lineno)
         return [raw_node]
 
-# TODO - I'm pretty sure this breaks for the case of multiple youtubes on one page.
-# we should push players to an array.
-# There should probably be only one function called onYouTubeIframeAPIReady
-# this function should iterate over all the various videos creating players
-#
+
 class Youtube(IframeVideo):
     """
 .. youtube:: YouTubeID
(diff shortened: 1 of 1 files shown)


runestone/video/video.py
```diff
@@ -214,37 +210,18 @@ class Youtube(IframeVideo):
    """
     html = '''
     <div class="runestone" style="margin-left: auto; margin-right:auto">
-    <div id="%(divid)s_vid"></div>
+	<iframe id="%(divid)s_vid" type="text/html" width="%(width)u" height="%(height)u"
+  src="https://www.youtube.com/embed/%(video_id)s?autoplay=0"
+  frameborder="0"></iframe>
     </div>
-    <script id="%(divid)s_script">
```
bnmnetp: Unless there is more to come on this, I can't accept it as is.   For grading and analytics purposes we need to be able to capture events.  That is why I went this route using the youtube api.  

I'm pretty usre the solution I outlined under the TODO will solve the problem and preserve our logging.

ascholerChemeketa: OK, I will look into implementing a version that preserves logging.

bnmnetp: Awesome!  Thanks

ascholerChemeketa:
Reworked the changes to preserve logging with multiple videos per page. Also supports start and end attributes for the youtube directives.

Trickiest bit is making sure that the youtube api script is loaded once per page after all the youtube divs are set up. Due to that timing issue, it can't be added via app.add_javascript. Current solution appends that script tag to the body once the page is loaded. If there is a better way to insert a script tag at the end of each page (ideally only pages that use the youtube directive), please let me know.

PR #668, opened 2018-09-09
Adding the ability to import json in the textbook

soney:
This PR adds the Skulpt JSON library to ActiveCode windows (run `import json` in an AC window).

It uses the json.sk library: https://github.com/trinketapp/json.sk

NOTE: The JSON.sk library has a few bugs in it that I will try to fix in a later PR

### DIFF
--- a/runestone/activecode/js/activecode.js
+++ b/runestone/activecode/js/activecode.js
@@ -88,6 +88,7 @@ ActiveCode.prototype.init = function(opts) {
         this.caption = ""
     }
     this.addCaption();
+    this.addJSONLibrary();
 
     if (this.autorun) {
         $(document).ready(this.runProg.bind(this));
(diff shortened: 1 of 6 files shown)


runestone/activecode/js/activecode.js
```diff
@@ -774,7 +775,16 @@ errorText.NotImplementedError = $.i18n("msg_activecode_not_implemented_error");
 errorText.NotImplementedErrorFix = $.i18n("msg_activecode_not_implemented_error_fix");
 
 
-
+ActiveCode.prototype.addJSONLibrary = function () {
+    Sk.externalLibraries = {
```
bnmnetp: This should check to see if there are already other externalLibraries specified elsewhere.  Otherwise this will wipe them out.

soney:
@bnmnetp Thanks; this should be fixed now. I also fixed some of the bugs that I was encountering with `json.dumps` and `json.loads` by packaging a new version of https://github.com/substack/json-stable-stringify

PR #702, opened 2018-11-12
added or{ on paired blocks

yynelson:
I added keyword "or" to the left of paired Parsons blocks in parsons problems. Some users of the books don't often realize that the purple blocks are prompting for "pick one from these two". So I added this feature of a "or" in between paired blocks by only changing the parsons.js file. It seems that this addition is much more indicative of the intended meaning for paired blocks.


bnmnetp:
Thanks @yynelson You are working with Barbara Ericson on this?

yynelson:
Hi, @bnmnetp , yes I am currently an undergraduate working with Barbara Ericson on this. She said that previously, users don't always realize that the purple blocks indicate that one should pick one out of the two highlighted blocks, and an "or" keyword would be more indicative of the intended use of paired blocks.

PR #714, opened 2018-12-05
Add autoversion

bjones1:
This provide a function in Sphinx to automatically add a version to JavaScript and CSS files. This helps to ensure that clients are running the correct version of the code in this repo.


bnmnetp:
This will have huge benefits.

- Should chromedriver.exe really be included in this?

runestone/common/runestonedirective.py
```diff
@@ -102,6 +108,63 @@ def _purge_runestone_data(app, env, docname):
...
+
+            # Append the CRC to the JavaScript reference.
+            return '{}?v={}'.format(filename, crc_str)
+
+    # No match was found.
+    raise ExtensionError('Unable to find {} in html_static_path.'.format(filename))
+
+
+# Convenience method to call ``add_javascript`` using ``add_autoversion``.
+def _add_autoversioned_javascript(self, filename):
```
bnmnetp: This is a great approach.

Seems like it would make a good PR for Sphinx itself.

bjones1: Thanks! I'll have to work on that (someday)...

bjones1:
Oops! chromedriver.exe is now removed. I rebased after several merges.

bjones1:
Thanks!

PR #718, opened 2018-12-10
Fix: Correctly JSON encode and decode fitb answers.

### LINKED ISSUE #156: Fill in the blank answers re-rendered incorrectly on page reload
Correct fill in the blank answers are being put in quotes and into a list when re-rerendered on page reload.

image

### LINKED ISSUE #716: Fill in the blank answers re-rendered incorrectly on page reload
Correct fill in the blank answers are being put in quotes and into a list when re-rerendered on page reload.

image

bjones1:
Hopefully fixes RunestoneInteractive/rs#156. Not tested, just a quick response...

### DIFF
--- a/runestone/fitb/js/fitb.js
+++ b/runestone/fitb/js/fitb.js
@@ -124,7 +124,13 @@ FITB.prototype.renderFITBFeedbackDiv = function () {
 
 FITB.prototype.restoreAnswers = function (data) {
     // Restore answers from storage retrieval done in RunestoneBase
-    var arr = data.answer.split(",");
+    try {
+        // The new encoding is JSON. Try that first.
+        var arr = JSON.parse(data.answer);
+    } catch (err) {
+        // Otherwise, fall back to the old comma encoding.
+        arr = data.answer.split(",");
+    }
     for (var i = 0; i < this.blankArray.length; i++) {
         $(this.blankArray[i]).attr("value", arr[i]);
     }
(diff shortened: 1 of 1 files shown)


runestone/fitb/js/fitb.js
```diff
@@ -124,7 +124,13 @@ FITB.prototype.renderFITBFeedbackDiv = function () {
 
 FITB.prototype.restoreAnswers = function (data) {
     // Restore answers from storage retrieval done in RunestoneBase
-    var arr = data.answer.split(",");
+    try {
+        // The new encoding is JSON. Try that first.
+        var arr = JSON.parse(data);
```
presnick: I get an error on this. But the following seemed to work.
```suggestion
        var arr = JSON.parse(data.answer);
```

bjones1: Oops! Thanks -- that's correct. I'll update my PR.

It looks like @bnmnetp already included part of this. but that code doesn't support old comma-style encoding. Brad, this seems like it would break when trying to restore old answers. Am I missing something?

bjones1: OK, sounds good.

bjones1:
Doesn't https://github.com/RunestoneInteractive/RunestoneComponents/blob/master/runestone/fitb/js/fitb.js#L154 also need to be updated to `var answer = JSON.stringify(storedData.answer)`?

PR #737, opened 2019-01-07
DRY: Get the local storage key from a single function

bjones1:
I somehow missed on DRY that should have gone in 4899df3c48141018ceb34bb6c7b060453b381ca9.


bjones1:
Thanks!

PR #742, opened 2019-01-11
added functions to extract text from a html format

dichha:


dichha:
Duplicate branch -- same as #738

PR #743, opened 2019-01-11
Fix: Prevent answers from "leaking" from one course to another

bjones1:
This PR includes the course name in the key used for local storage. Without this, a student who took a course last semester from a course named fopp_fall_2018 will already have answers provided if they re-take in this semester in fopp_spring_2019.

**However**, I assume this will break existing local storage. How big an issue is this? I could include code to check for the old label, but only if necessary.

### DIFF
--- a/runestone/common/js/runestonebase.js
+++ b/runestone/common/js/runestonebase.js
@@ -132,5 +132,5 @@ RunestoneBase.prototype.shouldUseServer = function (data) {
 
 // Return the key which to be used when accessing local storage.
 RunestoneBase.prototype.localStorageKey = function () {
-    return eBookConfig.email + ":" + this.divid + "-given";
+    return eBookConfig.email + ":" + eBookConfig.course + ":" + this.divid + "-given";
 }


bjones1:
That's what I'd expect. I just didn't want to go do this, only to find out it wasn't necessary. I'll work on it.

bjones1:
I looked at this, and fixing it will be hard, since each question type loads from local storage if the server isn't used. Fixing the server case is easier, however -- that's centralized in `runestone_base.js`.

Since any updates to the non-server case involves students getting another copy of the book, I think this is less an issue -- students probably get a copy at the beginning of the semester. What do you think about just providing a migration path for the server case, but not for the non-server case?

bnmnetp:
I think your solution is probably fine. Since local storage is browser and/or machine dependent.  There are plenty of lab situations in which students may log in, but their local storage or cache is wiped out by something like deep freeze anyway.

bjones1:
Hmm, this also runs into a dead end: fundamentally, all data is loaded in each question's `checkLocalStorage` function. However, the server will in almost all cases already have the answers stored -- it's only when the user saves an answer while offline before this fix, then comes back online after this fix that an answer won't be restored.

So, I'd propose merging this as is -- the benefits (not having answers from old courses automatically available in new courses) outweigh the losses (an unusual case where an answer won't be restored).

bnmnetp:
But now we have come full circle to my original problem with it, that it will break for all of the students in courses that are currently in progress.

I would actually really like to get this into production as I am seeing too many cases of students not getting credit for their work because even though they are registered for a course, they use Google to find their way back to the open instance of the same book.  They do their answers there but then they are not recorded in their course.  Once they login and find their way back to the proper page they are fooled because the answers still show from localstorage.

bjones1:
I don't see a perfect way forward. From what I can tell, this will not affect many users -- the only failing case I can think of is a user who works offline under the old code, then comes back on line after the update. In this case, they will lose their offline work. However, as you point out, students are already losing the offline work when they submit to the wrong class.

Sigh. Another approach would be to wait to the end of the semester, when the impact would be smaller.

???

bnmnetp:
I think I was just being stupid.  You are right that for the vast majority of the students this change won't have an impact because localstorage will fail, but their answers will be restored from the db.

I'll do a little more testing with this myself.  But I think its probably time to accept and do a new release of runestone.

bjones1:
Sounds good. This does close a longstanding bug and will be good going forward.

bjones1:
Pinging on this -- I don't want it to get lost in the busyness..

bjones1:
Thanks!

PR #772, opened 2019-02-19
bigger Parsons block numbered font-size

yynelson:
The css file has been modified to give the numbers on the parsons block a larger font-size. It also fixes the problem with code always floating to the left as it was before in APCSAReview 17.1.1. Parsons.js contains parsons block height changes that must go along with the change in numbered font-size in css.

### DIFF
--- a/runestone/parsons/js/parsons.js
+++ b/runestone/parsons/js/parsons.js
@@ -1438,7 +1438,7 @@ Parsons.prototype.initializeAreas = function(sourceBlocks, answerBlocks, options
 				addition = 2.1;
 			areaHeight += (item.outerHeight(true)+height_add*addition);
 			areaWidth = Math.max(areaWidth, item.outerWidth(true));
-			console.log(item.outerHeight(true));
+			
 		};
 	}
 	for (i = 0; i < blocks.length; i++) {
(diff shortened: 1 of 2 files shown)


ericsonga:
I have approved these changes.

runestone/parsons/js/parsons.js
```diff
@@ -1416,17 +1416,29 @@ Parsons.prototype.initializeAreas = function(sourceBlocks, answerBlocks, options
...
 	} else {
 		areaWidth = 0;
 		maxFunction = function(item) {
-			areaHeight += item.outerHeight(true);
+			var addition = 3.8;
+			if (item.outerHeight(true) != 38)
+				addition = 2.1;
+			areaHeight += (item.outerHeight(true)+height_add*addition);
 			areaWidth = Math.max(areaWidth, item.outerWidth(true));
+			console.log(item.outerHeight(true));
```
bnmnetp: Should remove this call to console.log

runestone/parsons/js/parsons.js
```diff
@@ -1435,6 +1447,7 @@ Parsons.prototype.initializeAreas = function(sourceBlocks, answerBlocks, options
 	this.areaWidth = areaWidth;
 	if(this.options.numbered != undefined) {
 		this.areaWidth += 25;
+		//areaHeight += (blocks.length);
```
bnmnetp: If its tested and working then this should be removed instead of commented out.

PR #805, opened 2019-05-01
Dynamic routing

bjones1:
This works with https://github.com/RunestoneInteractive/RunestoneServer/pull/1112 to implement dynamic book routing.

### DIFF
--- a/runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html
+++ b/runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html
@@ -132,7 +132,7 @@
 
         <li class="divider-vertical"></li>
 
-        {% if use_services == 'true' %}
+        {% if use_services %}
         <!-- user account dropdown -->
         <li class="dropdown">
           <a class="dropdown-toggle" href="#" data-toggle="dropdown">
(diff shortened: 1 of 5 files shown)


runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html
```diff
@@ -223,17 +223,28 @@
 
 <script type="text/javascript">
   eBookConfig = {};
-  eBookConfig.host = '{{course_url}}' ? '{{course_url}}' : 'http://127.0.0.1:8000';
-  eBookConfig.app = eBookConfig.host+'/{{appname}}';
+  {% if use_services %}
```
bnmnetp: @bjones1 -- The only hesitation I have about this is that it conflates using Runestone Ajax services with our move to dynamic pages.  I feel like it would be much more backward compatible to have a separate flag

bnmnetp: without something a bit more flexible then everyone who runs their own server will need to move to dynamic pages, they won't be able to rebuild any books to work in the old mode once this PR is accepted.

bnmnetp:
@bjones1 -- I merged this one with my changes.  

The biggest thing is that I added a separate dynamic_pages flag which made it easier to build static books that still wanted to use Runestone server.  I also made the database update backward compatible.

PR #816, opened 2019-05-28
Misc fixes

bjones1:
- A student was cut-and-pasting from a Word document, and had answers marked wrong in a fitb question. The updated fitb allows tabs (what the student pasted in) and other whitespace.
- The shortanswer question's encoding is fixed.
- mchoice formatting is now correct on Sphinx 2.0 (and still works on 1.8.x).

### DIFF
--- a/runestone/shortanswer/js/shortanswer.js
+++ b/runestone/shortanswer/js/shortanswer.js
@@ -134,7 +134,7 @@ ShortAnswer.prototype.submitJournal = function () {
 
 
     this.setLocalStorage({answer: value, timestamp: new Date()})
-    this.logBookEvent({'event': 'shortanswer', 'act': JSON.stringify(value), 'div_id': this.divid});
+    this.logBookEvent({'event': 'shortanswer', 'act': value, 'div_id': this.divid});
     this.feedbackDiv.innerHTML = "Your answer has been saved.";
     $(this.feedbackDiv).removeClass("alert-danger");
     $(this.feedbackDiv).addClass("alert alert-success");
--- a/runestone/unittest_base.py
+++ b/runestone/unittest_base.py
@@ -4,7 +4,6 @@ import sys
 import platform
 import subprocess
 from selenium import webdriver
-from selenium.webdriver.remote import webelement
 from pyvirtualdisplay import Display
 
 # Select an unused port for serving web pages to the test suite.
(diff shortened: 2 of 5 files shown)


runestone/unittest_base.py
```diff
@@ -41,7 +40,7 @@ def setUpModule(self):
         if self.exit_status_success:
             self.assertFalse(p.returncode)
         # Run the server. Simply calling ``runestone serve`` fails, since the process killed isn't the actual server, but probably a setuptools-created launcher.
-        self.runestone_server = subprocess.Popen(['python', '-m', 'runestone', 'serve', '--port', PORT])
+        self.runestone_server = subprocess.Popen([sys.executable, '-m', 'runestone', 'serve', '--port', PORT])
```
bnmnetp: I feel like we are going back and forth here in that sys.executable doesn't work when the system thinks executable is runestone.  I thought I had to change this back to python in this exact same place.

runestone/shortanswer/js/shortanswer.js
```diff
@@ -134,7 +134,7 @@ ShortAnswer.prototype.submitJournal = function () {
 
 
     this.setLocalStorage({answer: value, timestamp: new Date()})
-    this.logBookEvent({'event': 'shortanswer', 'act': JSON.stringify(value), 'div_id': this.divid});
+    this.logBookEvent({'event': 'shortanswer', 'act': value, 'div_id': this.divid});
```
bnmnetp: Why remove JSON.stringify ?

runestone/unittest_base.py
bjones1: I don't see it -- the original commit of 63bf26e4cbd3dd771e69544e35bde94678979f03 was `self.runestone_server = subprocess.Popen(['python', '-m', 'runestone', 'serve', '--port', PORT])`. I do remember this on the server, where NGINX's emperor mode broke `sys.executable` when trying to do a runestone build for a problem written by the instructor in the admin interface.

runestone/shortanswer/js/shortanswer.js
bjones1: Because the server doesn't JSON decode it. Hence, there's an extra level of JSON introduced, which this commit fixes.

It doesn't need to be JSON encoded then decoded, since logBookEvent already does this.

runestone/unittest_base.py
bnmnetp: Yes, that was the case I was thinking of.  So in my mind that makes sys.executable unreliable.

bjones1: I see you merged this, so I'm assuming you're OK with this.

For the record, from https://docs.python.org/3/library/sys.html#sys.executable is "A string giving the absolute path of the executable binary for the Python interpreter, on systems where this makes sense. If Python is unable to retrieve the real path to its executable, sys.executable will be an empty string or None."

In the case of testing, we can assume CPython, hence a correct `sys.executable`.

PR #817, opened 2019-05-29
Literate programming additions

bjones1:
This PR provides literate programming capabilities, when paired with a soon-to-be submitter PR for the server.


bjones1:
WIP -- I haven't been able to fix the test failure yet. It runs fine on Windows.

bnmnetp:
Well, it doesn't seem like it should be the cause of the problem, but I'm getting a 404 on html4css1.css when the test page tries to load.  I think it might be a missing requirement from CodeChat.css

bnmnetp:
Also this PR has definitely broken some formatting

Master with correct width for parsons container:
Screen Shot 2019-05-29 at 7 38 32 AM

PR
Screen Shot 2019-05-29 at 7 39 03 AM

bjones1:
Thanks! My CSS changes made the test fail (arrgh, that's subtle). Thanks for pointing that out. I've backed all but the essential CSS changes out (styling only for LP stuff).

This is now ready to go.

PR #818, opened 2019-05-30
server side reveal

bnmnetp:
This implements a feature whereby the contents of an instructoronly reveal are only sent to the browser if the current user is an instructor.

### DIFF
--- a/runestone/reveal/reveal.py
+++ b/runestone/reveal/reveal.py
@@ -48,22 +48,35 @@ def visit_reveal_node(self, node):
     else:
         node.reveal_options['modaltitle'] = ''
 
-    res = TEMPLATE_START % node.reveal_options
+    if node.reveal_options['instructoronly'] and node.reveal_options['is_dynamic']:
+        res = DYNAMIC_PREFIX
+    else:
+        res = ""
+
+    res += TEMPLATE_START % node.reveal_options
     self.body.append(res)
 
 def depart_reveal_node(self,node):
 #Set options and format templates accordingly
     res = TEMPLATE_END % node.reveal_options
+    if node.reveal_options['instructoronly'] and node.reveal_options['is_dynamic']:
+        res += DYNAMIC_SUFFIX
 
     self.body.append(res)
 
 #Templates to be formatted by node options
+DYNAMIC_PREFIX = '''
+{{ if is_instructor: }}
+'''
 TEMPLATE_START = '''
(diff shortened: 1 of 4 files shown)


bnmnetp:
See #818 RunestoneServer/1243

runestone/reveal/reveal.py
```diff
@@ -48,22 +48,35 @@ def visit_reveal_node(self, node):
...
+    else:
+        res = ""
+
+    res += TEMPLATE_START % node.reveal_options
     self.body.append(res)
 
 def depart_reveal_node(self,node):
 #Set options and format templates accordingly
     res = TEMPLATE_END % node.reveal_options
+    if node.reveal_options['instructoronly'] and node.reveal_options['is_dynamic']:
```
bjones1: This is repeated twice -- you might want to assign it to a variables.

runestone/reveal/test/pavement.py
```diff
@@ -32,7 +32,8 @@
                        'use_services': 'false',
                        'python3': 'false',
                        'dburl': '',
-                       'basecourse': 'reveal-test'
+                       'basecourse': 'reveal-test',
+                       'dynamic_pages': True
```
bjones1: It would be nice to have two test cases, one with dynamic and other without.

bnmnetp: Ugh!

bjones1: Yes, that's a pain.

**But**, if we actually commit to two cases (static and dynamic), then part of the additional burden is writing those cases. (Not that I did this for the original dynamic book routing PR...ahem...). Fundamentally, supporting both static and dynamic versions of the site will be a lot of work, especially if we implement a lot of neat dynamic features.

runestone/reveal/reveal.py
bnmnetp: Repeated but in two different functions.  It seems more clear about what it is we are doing.

runestone/reveal/test/pavement.py
bnmnetp: I think if we want to test all of these for both dynamic_pages == True / False then we should find a way to parameterize the pavement.py file at run time.  I think that would be a reasonable thing to do in a second PR so that books can be built multiple times with different settings.  

Maybe we need to update the runestone build command to allow for any of the template variables to be overwritten from the command line.

Also, this is not really a very good test of the dynamic mode as all it does is verify that the template if statement is there.  To really test the dynamic mode stuff we ought to do in the context of running in a docker container with a fully capable server.

bjones1: Good points -- I agree!

PR #822, opened 2019-06-13
Clean: Use the _static/ folder from Runestone Components

bjones1:
instead of copying it to each book.


runestone/__init__.py
```diff
@@ -31,7 +31,11 @@ def runestone_static_dirs():
     module_static_image = ['%s/images' % os.path.join(basedir,x) for x in module_paths if os.path.exists('%s/images' % os.path.join(basedir,x))]
     module_static_bootstrap = ['%s/bootstrap' % os.path.join(basedir,x) for x in module_paths if os.path.exists('%s/bootstrap' % os.path.join(basedir,x))]
 
-    return module_static_js + module_static_css + module_static_image + module_static_bootstrap
+    return (
+        module_static_js + module_static_css + module_static_image +
+        module_static_bootstrap +
+        [os.path.join(basedir, 'common/project_template/_static')]
```
bnmnetp: I'm not 100% sure, but this may be a place where you should be using pkg_resources like I did in my PR to get the path.  This probably works in the dev environment, but I'm not too confident about it working from a pip install.

bjones1: I can, but the remainder of the function doesn't, so I was following the style:

``` Python
def runestone_static_dirs():
    basedir = os.path.dirname(__file__)
    module_paths = [ x for x in os.listdir(basedir) if os.path.isdir(os.path.join(basedir,x))]
    module_static_js = ['%s/js' % os.path.join(basedir,x) for x in module_paths if os.path.exists('%s/js' % os.path.join(basedir,x))]
    module_static_css = ['%s/css' % os.path.join(basedir,x) for x in module_paths if os.path.exists('%s/css' % os.path.join(basedir,x))]
    module_static_image = ['%s/images' % os.path.join(basedir,x) for x in module_paths if os.path.exists('%s/images' % os.path.join(basedir,x))]
    module_static_bootstrap = ['%s/bootstrap' % os.path.join(basedir,x) for x in module_paths if os.path.exists('%s/bootstrap' % os.path.join(basedir,x))]

    return (
        module_static_js + module_static_css + module_static_image +
        module_static_bootstrap +
        [os.path.join(basedir, 'common/project_template/_static')]
    )
```

bnmnetp: Ah, I guess it has the same effect then because we are getting `basedir` from `__file__` If its working for the other stuff I guess it will work fine.

PR #823, opened 2019-06-14
Login updates

bnmnetp:
Clean: -- removed a bunch of code from bookfuncs.js After this is merged, sites that are using the old style non-dynamically-served pages, will need to use a released version of  runestone < 4.0

Sites using the master branch should move to dynamic pages.

The big benefit of this PR is it elminates 3 or 4 round trip API calls that happened on every page load!

This PR is fully supported by any version of runestone server after the 4.0dev tag on RunestoneServer

@presnick tagging you because I know you tend to use the github repo instead of released versions.


bnmnetp:
@bjones1 -- These changes seem to have broken The last two lines of the lp tests.  I'm not sure how this is supposed ever work without a server...

bjones1:
Great work cleaning up. I'll be happy to fix the LP tests when I'm back at my desk.

PR #830, opened 2019-06-25
Fix tests

bjones1:
Misc test fixes/cleanup.

### DIFF
--- a/runestone/activecode/test/test_activecode.py
+++ b/runestone/activecode/test/test_activecode.py
@@ -90,6 +90,8 @@ class ActiveCodeTests(RunestoneTestCase):
         rb.click()
         pb = self.driver.find_element_by_id("subchapterprogress")
         self.assertIsNotNone(pb)
+        # Wait for the JS to run. Increase this delay if the next assertion fails.
+        time.sleep(1)
         total = self.driver.find_element_by_id("scprogresstotal").text.strip()
         self.assertEqual(2, int(total))
         possible = self.driver.find_element_by_id("scprogressposs").text.strip()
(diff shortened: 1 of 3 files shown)


runestone/activecode/test/test_activecode.py
```diff
@@ -90,6 +90,8 @@ def test_activity_count(self):
         rb.click()
         pb = self.driver.find_element_by_id("subchapterprogress")
         self.assertIsNotNone(pb)
+        # Wait for the JS to run. Increase this delay if the next assertion fails.
+        time.sleep(1)
```
bnmnetp: I assume this is what was causing the failures on your windows machine?

Just for posterity, I looked a bit at https://selenium-python.readthedocs.io/waits.html  I don't think that provides a better solution than time.sleep but looking at many other areas in our tests there are definitely places where using the wait mechanisms would be much better than time.sleep.

bjones1: Yes, this fixes Windows tests.

Re waiting -- good point. My guess (untested):

``` Python
from selenium.webdriver.common.by import By
from selenium.webdriver.support.ui import WebDriverWait
from selenium.webdriver.support import expected_conditions as EC

        total = WebDriverWait(self.driver, 1).until(
            EC.text_to_be_present_in_element((By.ID, "scprogresstotal"), "2")
        assert total.text == "2"
```

bnmnetp: Yeah I saw that, but I assumed (without reading any docs) that since there was already text present in that element that it would not help us.  Maybe I'll read 😜

PR #862, opened 2019-07-29
Fix: Update old links to static books.

bjones1:


bjones1:
I've updated this to include changes made to master since I submitted this PR.

runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html
```diff
@@ -247,7 +251,7 @@
     {% endraw %}
   {% else %}
     eBookConfig.useRunestoneServices = {% if use_services == 'True' -%}true{%- else -%}false{%- endif -%};
-    eBookConfig.host = '{{course_url}}' ? '{{course_url}}' : 'http://127.0.0.1:8000';
+    eBookConfig.host = '{{course_url}}' || 'http://127.0.0.1:8000';
```
bnmnetp: Why make this change which now conflicts?  There was nothing broken about the code that was in place.

bjones1: Just trying to DRY the code a bit. I can rebase this and omit this if you'd like.

bnmnetp: I think its more of a philosophical difference.

As the reviewer in chief I prefer more of a "If it ain't broke don't fix it" style, because it keeps the PRs small and focused on what they say they are doing, and easier for me to figure out.

In this PR a small percentage of the changes make necessary changes to the links to old static books.  The rest of it leaves me puzzling over why you changed something that was working fine.

bjones1: Would you prefer me to re-factor this PR into two (or more) separate PR, or provide some explanation of what I'm doing?

bnmnetp: I don't think you need to break this one up as I've already spent time trying to figure it out. 

Some explanation of why you do the whole:

```
{% if dynamic_pages == 'True' %}
  {% set appname = '{{ =request.application }}' %}
{% endif %}
```

Even after thinking about it, its not obvious to me if this will even work when dynamic_pages is not true, or even why it is necessary at all.

bjones1: OK. Sorry for the confusion. I'll try to break PRs into smaller pieces.

If `dynamic_pages` isn't true, then nothing changes -- the code has no effect. It therefore relies on the setting of `appname` in `pavement.py`. When `dynamic_pages` is true, this fixes the value by pulling from the server -- assuming `runestone` in all cases is not correct. (I run a `runestone` application, then a slightly modified `exam_runestone` for testing.)

runestone/common/project_template/_templates/plugin_layouts/sphinx_bootstrap/layout.html
```diff
@@ -114,11 +118,11 @@
           </a>
           <ul class='dropdown-menu'>
             {% if 'overview' in course_id %}
-                <li><a href='/{{appname}}/static/{{course_id}}/overview.html' aria-label="index-page">Table of Contents</a></li>
+                <li><a href='{{ pathto(master_doc) }}/overview.html' aria-label="index-page">Table of Contents</a></li>
             {% else: %}
                 <li><a href='{{ pathto(master_doc) }}' aria-label="index-page">Table of Contents</a></li>
             {% endif %}
-            <li><a href='/{{appname}}/static/{{course_id}}/genindex.html'>Book Index</a></li>
+            <li><a href='{{ pathto(master_doc) }}/genindex.html'>Book Index</a></li>
```
bnmnetp: This produces incorrect paths...

For example from index.html you get
```
<li><a href='#/genindex.html'>Book Index</a></li>
```


From another file I get:
```
<li><a href='index.html/genindex.html'>Book Index</a></li>
```

bnmnetp: I think we want to use `pathto("genindex.html", 1)` to give us a relative path to the file itself.

bjones1: Oops! Thanks, that's correct -- I've tested it on my install after a book rebuild.

PR #876, opened 2019-08-13
Fix: Fully specify import of componentdb.

bjones1:
In general, starred imports are a bad thing. Here's one small step toward removing them -- fully specify an import for `get_dburl` instead of allowing a starred import to do the job. Ideally, the same update would be applied to all the `pavement.py` files out there as well.

### DIFF
--- a/runestone/activecode/test/pavement.py
+++ b/runestone/activecode/test/pavement.py
@@ -3,7 +3,7 @@ from paver.easy import *
 import paver.setuputils
 paver.setuputils.install_distutils_tasks()
 import os, sys
-from runestone.server import get_dburl
+from runestone.server.componentdb import get_dburl
 from sphinxcontrib import paverutils
 
 sys.path.append(os.getcwd())
--- a/runestone/assess/test/pavement.py
+++ b/runestone/assess/test/pavement.py
@@ -3,7 +3,7 @@ from paver.easy import *
 import paver.setuputils
 paver.setuputils.install_distutils_tasks()
 import os, sys
-from runestone.server import get_dburl
+from runestone.server.componentdb import get_dburl
 from sphinxcontrib import paverutils
 import pkg_resources
 
--- a/runestone/clickableArea/test/pavement.py
+++ b/runestone/clickableArea/test/pavement.py
@@ -3,7 +3,7 @@ from paver.easy import *
 import paver.setuputils
 paver.setuputils.install_distutils_tasks()
 import os, sys
-from runestone.server import get_dburl
+from runestone.server.componentdb import get_dburl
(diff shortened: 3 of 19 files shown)


bnmnetp:
Genuinely curious:  Was there some kind of bug that led you to make this PR?

My understanding is that this use of import * is fairly common in creating packages where you don't necessarily want to expose the underlying Python files below a certain level in your package.  It does not pollute the global namespace, it only imports * into the runestone.server namespace.  And if the package developer is careful and responsible that doesn't really hurt anyone.

bjones1:
This is a side-trail of noticing that we import too much stuff on the server from the components. Specifically, the server does an `from runestone import cmap` in `admin.py`, and a `from runestone.server import lp_common_lib` in `modules/feedback.py`. Which means we're (due to starred imports) actually creating a connection to the database every time `admin.py` runs. Yuck!

So, I started looking for ways to avoid the "import everything" approach, and this was one result.

bjones1:
You have good points -- I'm closing this, since I agree that the existing design is fine. Sorry for the noise.

bnmnetp:
I think needlessly creating yet another database connection is also problematic.  I think we ought to put all of the database globals inside an object and only initialize that object before its first use.

bjones1:
I agree! I also think that importing all of `runestone` from the server is inefficient and should be addressed.

bnmnetp:
I only see 2 places where we import from runestone

We import cmap in admin.py 

We import STUDENT_SOURCE_PATH and other stuff in feedback.py

I think both would be speedier if we only initialize database connections in server.py when needed.

These imports don't affect Ajax.py or books.py which are the most time critical.

bjones1:
I agree that this is somewhat low-priority, since it's not in the critical path (books or ajax). I think it brings up a wider issue: IMHO, we need some way for both the server and the components to import from a shared place without requiring a bunch of unrelated imports., rather than messing with db connections. Currently, `runestone.__init__` does these imports, which prevents us from using `runestone` as a shared area with some refactoring.

bnmnetp:
1.  It may be that with some refactoring runestone could be fine.  I'm not at all sure that all of those import * statements are necessary, but changing that would be a huge change to the way others view and use the API.  In terms of the __init__ file itself all we really need to import are the directive classes.

2.  Initializing the database stuff on demand is a problem we can solve.  That eliminates the database connection issue.

3.  Let's make an issue that addresses a specific need so we can discuss and design a solution.  Perhaps there is come kind of other package that sits in the middle between runestone components and the server.  What are the needs?  what are the use cases?

bjones1:
OK, done is RunestoneInteractive/rs#110.

PR #898, opened 2019-08-30
p5js directive added

schellenberg:
Adds a [p5js](https://p5js.org/) directive, using [p5.js-widget](https://github.com/toolness/p5.js-widget). An example of the directive is:

`.. p5:: first-demo
    :width: 400
    :height: 400
    :autoplay:

    function setup() {
      createCanvas(350, 300);
    }

    function draw() {
      background(255);
      fill(0);
      ellipse(mouseX, mouseY, 50, 50);
    }`


bnmnetp:
Thanks for the PR!  What a cool surprise after the long weekend.

Could you add a test folder under runestone/p5 that includes a one page "book" that demonstrates the use of the p5 directive?   Ideally this would also include a simple Selenium unit test, that can come a bit later if you need some help getting up to speed on Selenium.

Did you consider adding this as a sub-class of the activecode directive?   That would have been my suggestion, as I did with Javascript, HTML, SQL, and the non-browser languages) but maybe there were challenges that made that harder.

schellenberg:
I added a simple one-pager that shows the options of the p5 directive. I've not used Selenium before, and don't have the time to explore it right now, so there isn't a unit test.

I hadn't thought of adding p5 as a sub-class of the activecode directive. That might be a really nice way to go. To be honest, I was just hacking something together to use for a class that I'm teaching, and went with what I thought would be the path of least resistance...

bnmnetp:
I just heard from another person today that is interested in using processing on Runestone.

Our old version -- some variant of pyprocessing -- is suffering from a bad case of bit rot and no longer works.  Not surprising as I haven't given it any love since 2013.

I would love to see this accepted -- especially if we could make it another subclass of the activecode class -- ala SQL, Javascript, HTML and others.

I'd love to see someone take up the challenge so I could get it accepted!

PR #920, opened 2019-09-18
Add Windows code to kill a stuck server.

bjones1:


runestone/unittest_base.py
```diff
@@ -36,9 +37,14 @@
...
 PORT = "8081"
 # Use the localhost for testing.
-HOST = "http://127.0.0.1:" + PORT
+HOST_ADDRESS = "127.0.0.1:" + PORT
+HOST_URL = "http://" + HOST_ADDRESS
 
 
+# Define the platform.
+IS_WINDOWS = platform.system() == "Windows"
+IS_LINUX = sys.platform.startswith("linux")
```
bnmnetp: could you make this linux or darwin ? this should cover the macOS folks with out any problems.

bjones1: Are you sure? I use this only to determine if we should run pyvirtualdisplay, and I can't tell if mac supports that or not. See line 149.

bnmnetp: I was just trying to make sure that we didn't include windows by excluding macOS.  the pyvirtualdisplay issue probably doesn't matter for Mac.

bjones1: Since the only use of `IS_LINUX` is for pyvirualdisplay, I'd say leave this as it is. Sound good?

PR #932, opened 2019-10-05
Clarify difference between RunestoneServer and `runestone serve`

### LINKED ISSUE #931: Does local development require a database?
I'm trying to contribute to this project for the first time. I followed the instructions in the README and things went smoothly. However when I run any `runestone` command the first thing I see is:

```
Cannot configure a Database URL!
Skipping all DB operations because environment variables not set up
```

Digging around in the source code I figured out I could set `WEB2PY_CONFIG` and `DEV_DBURL`. I tried sqlite and that failed with an obscure error, so I tried postgres and that sort of worked, after installing psycopg2 and making a database. Except it didn't actually work because none of the tables exist.

Do I need a database for testing and development? If not, I think there should be some documentation or an error message indicating this. If so, how is it set up?

alexmojaki:
Based on the discussion in https://github.com/RunestoneInteractive/fopp/issues/269. Closes #931.

Maybe it's just me, but when I read the README, I interpreted "the builtin Python webserver" to mean something like `python -m http.server`. I didn't mentally link it to the command beneath. Hopefully a colon and moving that warning about six out of the way fixes that.

But I've put the main clarification in the command because one should generally assume that people don't read the docs.

There's probably more that can be done in this area. There's so many repositories and pages and it's hard to see how it all fits together. Further down it says:

> (See the RunestoneServer repository and http://runestoneinteractive.org for more complete documentation on how this project works.)

but the first thing I see then is the blog, and I don't know which page I'm meant to be opening to see the documentation it's referring to.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,14 +77,11 @@ To build the included default project run
 
     runestone build
 
-You will now have a build folder with a file index.html in it, along with some default content.  The contents of the build folder are suitable for hosting anywhere that you can serve static web content from!  For a small class you could even serve the content using the builtin Python webserver.
-
 *Note:* If you come across version conflict with ``six`` library while building the project, ``pip install --ignore-installed six`` command might be useful.
 
-::
+You will now have a build folder with a file index.html in it, along with some default content.  The contents of the build folder are suitable for hosting anywhere that you can serve static web content from!  For a small class you could even serve the content using the provided Python webserver::
 
     $ runestone serve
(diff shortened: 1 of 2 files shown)


bnmnetp:
Thanks, its always great to have someone new who is willing to point out (and fix!) our blind spots, hidden assumptions, and things that we gloss over because we've known about them for years.

Issue #864, opened 2019-08-02
Firefox displays next/previous page arrows with white background


cigas:
In Firefox, the next/previous arrows are not in a transparent box, so they can occlude text, especially when using a larger font. I don't see this issue in Chrome.

![image](https://user-images.githubusercontent.com/11619455/62335588-ecf13880-b491-11e9-850b-1fe4a3765285.png)

adejumool:
@bnmnetp Would it be good solution to making the arrows transparent or invisible until a hovered on.

-Tobi Adejumo, Berea College Student

bnmnetp:
I think transparency would be a good compromise.  I'm not in favor of invisibility 😄

PR #959, opened 2019-11-04
Issue #864: Fixed Nav Buttons

NateJose123:
Made the nav buttons transparent till they are hovered on.

-Aaron, Tobi and Nathan | Berea College Students


bnmnetp:
Good work thanks for getting this done.

One question I have for you as I was reviewing the code.  Did you investigate using a CSS :hover rule rather than writing the Javascript functions?  Either way works, just curious.

Issue #953, opened 2019-10-31
Addition of C++ syntax highlighting


cbondurant:
I am one of the members of the team who are working on adding color syntax support through codemirror for C++. 
We have currently gotten so far as to have basic highlighting, as in comments, number literals, and string literals. Still, we believe that the updating of the codemirror files within runestone components will resolve this. 
There are additional complications in the fact that codemirror is passed its language mode through the same variable used to decide what kind of activecode runner will be used. `cpp` is used in the RST file for the language variable. However, codemirror expects c syntax highlighting to be referred to by `clike`. We are currently unsure where precisely this value is passed to the actual rendered webpage, and expect that it will need to be resolved before this can be completed.

bnmnetp:
Look at the `createEditor` function in `activecode.js` you will see how I handled for setting up java and sql.

cbondurant:
closing issue as this was merged over half a year ago.

PR #960, opened 2019-11-04
Update Codemirror to ver. 5.49.2 with support for C++ syntax highlighting

cbondurant:
This pull request resolves issue #953 
Activecode was updated to change the cpp mime type to `text/x-c++src`, allowing codemirror to properly recognize the cpp type. This allowed the books to color-code the cpp activecode, which is now displaying functionally.

We believe this is working without issue, but we cannot verify that the activecode for c++ will actually build, due to our testing environment (image below).
Capture

Since the codemirror plugin updated through several versions, it is likely that there are other features that have been added to the plugin that we have not fully explored. 

We are berea college students.

### DIFF
--- a/runestone/activecode/js/activecode.js
+++ b/runestone/activecode/js/activecode.js
@@ -127,6 +127,8 @@ ActiveCode.prototype.createEditor = function (index) {
         edmode = 'text/x-sql'
     } else if (edmode === 'java') {
         edmode = 'text/x-java'
+    } else if (edmode === 'cpp') {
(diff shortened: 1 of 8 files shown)


bnmnetp:
Thanks, this is exciting.

I'm going to pull this onto my development server so I can play use it for a few days before I merge.

bnmnetp:
Everything seems to be working fine.  All languages are colored correctly and all run as they should.

I'll likely include this in the next Saturday release.

PR #965, opened 2019-11-12
do not specify a version of urllib

bnmnetp:
Use sphinx > 2.0 and leave off urllib3

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,9 @@
 click
 Paver>=1.2.4
 six
-Sphinx>=1.8.5
+Sphinx>=2.0.0
 sphinxcontrib-paverutils>=1.17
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,10 +1,9 @@
...
 Paver>=1.2.4
 six
-Sphinx>=1.8.5
+Sphinx>=2.0.0
 sphinxcontrib-paverutils>=1.17
 cogapp>=2.5
 SQLAlchemy>=1.0.13
-urllib3==1.24.2
 # CodeChat is Python 3 only.
 CodeChat; python_version != '2.7'
```
bjones1: I'd suggest removing the condition, since we're now Python 3 only.

Issue #958, opened 2019-11-04
WCAG Compliance 4.1.2 Name, Role, Value


EzraSkwarka:
We are verifying and improving Runestone's level of WCAG compliance to Level AA, our findings are summarized on a Google Doc [here](https://docs.google.com/document/d/1XKiq4WanGln44by62XkSvAqjpse_mElYNoYtCIvtgBk/edit?usp=sharing). _This is for guideline 4.1.2_

>**4.1.2 Name, Role, Value**: For all user interface components (including but not limited to: form elements, links and components generated by scripts), the name and role can be programmatically determined; states, properties, and values that can be set by the user can be programmatically set; and notification of changes to these items is available to user agents, including assistive technologies. (Level A)

The only issue we could find with this topic is that for the arrows between chapters we need to list where they go _before_ they are hovered over.

Part of the Berea College Open Source Class @clarkh97 @owenslo @EzraSkwarka

EzraSkwarka:
The arrows also violate 1.3.3 
>**1.3.3 Sensory Characteristics:** Instructions provided for understanding and operating content do not rely solely on sensory characteristics of components such as shape, size, visual location, orientation, or sound. (Level A)
Note: For requirements related to color, refer to Guideline 1.4.

PR #969, opened 2019-11-20
Fixed issue #958 

NateJose123:
Made arrows transparent when window is resized and added text under arrows for accessibility.

- Aaron, Tobi and Nathan | Berea College Students


NateJose123:
Hello @bnmnetp,

Aaron, Tobi and I were wondering why the build failed as indicated by the compiler above. We looked through the log and found that the errors were not related to any of the files we worked on and our build ran just fine on our virtual environment. Please could you advise regarding this.

Thank you.

bnmnetp:
It looks like it is an upstream dependency conflict that is causing the build to fail.  This happens occasionally.

I'll look into updating the requirements.txt file to eliminate the problem.

PR #972, opened 2019-12-02
Closes #947. Created dark theme.

natstall:
Added an optional dark theme with a toggle in a dropdown from the navbar. - Berea student. With @mayjue.


bnmnetp:
Really nice work team!

I have a couple of minor comments that I would like to have you work on before I accept and deploy this.

1.  I would prefer to have the switch to go to dark mode under the 👤 menu .  We have so many menus already and having a menu with just a single item under it takes up a lot of space.

2. The contrast between the ✖️and the red background, and the ✔️and the purple background is not very good -- at least to my older eyes.  I would like to see you either make the red/purple a little softer or make the color of the symbols bolder.  I lean a bit toward the toning down the red/purple, but that may just be my personal preference.   It is important for us to maintain a good contrast for people with impaired vision.

Thanks again for the great work on this.

natstall:
I've moved the toggle and softened the red and purple alerts as requested. @mayjue

Issue #991, opened 2020-03-12
Dark mode update?


GregDowl:
Personally I don't really like the look of the current dark mode, I've tried finding where it is in the code but am not having any luck getting it to work locally.

If anyone can make a more viewable dark mode that would be amazing

bnmnetp:
What is it that you don't like?  Web site appearances are pretty subjective, so just vaguely stating that you don't like something isn't all that useful.

For some context, this feature was added by a group of students taking part in a course on contributing to open source software projects.  See PR #972 to see the files they changed.

PR #992, opened 2020-03-12
potential dark mode update

neeeeel:
fixes issue #991 

I would greatly appreciate feedback on the dark mode color scheme i've chosen


bnmnetp:
A screenshot would be pretty useful to avoid having to merge this and build a book just to see it.

In any case, I'm not sure how to judge on theme from another, as this is pretty subjective.

What problem are you trying to solve with this PR?

neeeeel:
So personally I agree with the creator of issue #991, the live dark mode is rather harsh on the eyes, with the updates I've proposed I think makes the dark mode a lot more usable for a longer period of time. Currently, I'm not at the system in which I can show the changes but once I can, I'll put up a screenshot.
If you don't find it beneficial, that's completely fine because like you said, it is almost purely subjective.

neeeeel:
Here's the screenshot.

What's on the left is the live dark mode and what's on the right is my change. In my opinion, the live dark mode is too harsh in it's colors and it almost seems like it's just the opposite color of the regular mode which shouldn't be how a dark mode looks. You are 100% right, it is completely subjective to each user but I think it is an improvement. 

Screen Shot 2020-03-13 at 2 31 11 PM

bnmnetp:
Does your PR address RunestoneInteractive/RunestoneServer#1439 ?  That would certainly be an enticement for me to accept this. 😄

bnmnetp:
Thanks, I've incorporated your changes into some recent work.

PR #1002, opened 2020-04-30
Misc fixes

bjones1:
- Update the literate programming component with a minor bugfix.
- Update to CodeChat 1.8.0, a dependency of the literate programming component.
- Address deprecation warnings in Sphinx 3.0.


bjones1:
OK, thanks -- updated.

bjones1:
I've just pushed changes after rebasing onto the post-webpack master. I've fixed several more items that I found as well.

runestone/common/runestonedirective.py
```diff
@@ -337,11 +337,11 @@ def add_skulpt_js(app):
     if first_time(app, "add_skulpt_js"):
         app.add_autoversioned_javascript("skulpt.min.js")
         app.add_autoversioned_javascript("skulpt-stdlib.js")
-        app.add_javascript("https://cdn.jsdelivr.net/npm/vega@4.0.0-rc.2/build/vega.js")
-        app.add_javascript(
+        app.add_js_file("https://cdn.jsdelivr.net/npm/vega@4.0.0-rc.2/build/vega.js")
```
bnmnetp, line 340: As skulpt is now bundled into runestone.js this probably isn't necessary.

bjones1: OK, great! I assume this means I can delete lines 340-346 from the PR, then?

bnmnetp: yes

runestone/__init__.py
```diff
@@ -41,6 +41,7 @@ def runestone_static_dirs():
     module_static_css.append(os.path.join(basedir, "accessibility", "css"))
     module_static_css.append(os.path.join(basedir, "webgldemo", "css"))
     module_static_css.append(os.path.join(basedir, "matrixeq", "css"))
+    module_static_css.append(os.path.join(basedir, "lp", "css"))
```
bnmnetp, line 44: I need to think this through a bit...  whatever we add to static dirs means that sphinx will copy the files from the path to the static folder... mostly this is redundant as nearly all of the files in the css/js folders for a component are bundled and do not need to be copied.

bjones1: True -- but in this case, https://github.com/bjones1/CodeChat/blob/master/CodeChat/css/CodeChat.css#L31 performs the include. This file isn't processed by webpack, so it needs to be saved as a standalone file for inclusion.

runestone/common/runestonedirective.py
bjones1: OK, done in d867c61b549fb6e59f101bac6b06ffd84012fadd.

runestone/__init__.py
bnmnetp: But is there some reason why we can’t include it in webpack?  The goal is to minimize the number of individual scripts and css links when they can be just as easily imported by webpack.

bjones1: I don't know. I see an error at https://github.com/bjones1/CodeChat/blob/master/CodeChat/css/CodeChat.css#L31 when that imports fails, so I'm assuming it doesn't work...

That is, can CSS not included in the webpacked file still do @imports as usual? I don't think so...

bnmnetp: I have css-loader configured, but I think I failed to enable it for css, just for png,jpg,svg...  https://webpack.js.org/loaders/css-loader/

Just add css to line 46 of webpack.config.js and see if that does it.

PR #1011, opened 2020-06-17
Modified the parsons test case and updated parsons.js to use localStorageKey()

Coldstream-Louis:
test_parsons.py: 
1. Changed the code in test_help() function to make the test pass without providing indentation, and combined two correct blocks which have the max distance(If there is one correct block not in the answer area, this block will be picked and combined first). 
2. Added a function named test_indentation(), to test there is no indentation provided in intra-problem adaption when clicking the "Help me" button. 

parsons.js: 
Use localStorageKey() to generate the key from local storage. See line 1083-1089, and line 2257-2262


package.json
```diff
@@ -26,6 +26,7 @@
         "webpack-dev-server": "^3.10.1"
     },
     "dependencies": {
+        "-": "0.0.1",
```
bnmnetp, line 29: What is this? "-" ??  Is it necessary for parsons?

runestone/parsons/js/parsons.js
```diff
@@ -2250,12 +2253,13 @@ export default class Parsons extends RunestoneBase {
     //    checkCount = how many times it has been checked before correct
     //    userRating = 0..100 how good the person is at solving problems
     initializeAdaptive() {
-        this.adaptiveId = eBookConfig.email;
-        if (this.adaptiveId == undefined) {
-            this.adaptiveId = "parsons";
-        } else {
-            this.adaptiveId = this.adaptiveId + "-parsons";
-        }
+        // this.adaptiveId = eBookConfig.email;
```
bnmnetp, line 2256: When the unit tests pass, you will need to remove these lines before I accept the PR.   When you are using git there is no reason to leave commented out code lying around as you can always check out previous versions of the code.  Thanks.

package.json
Coldstream-Louis: Oh, I didn't make any change in package.json.

runestone/parsons/js/parsons.js
Coldstream-Louis: Sure I will remove it. Sorry for the incovenience

Coldstream-Louis:
Sorry I didn't update to the new version of the runestone that day. So I have got a little problem with the parsons problem not working. Now I fixed it. I will check the code before I pull a new request.

bnmnetp:
You don't need to make a new pull request.  If you push your changes to your Fork on github this pull request will automatically update.

runestone/parsons/js/parsons.js
bnmnetp: No worries, this is what the review process is for.

PR #1017, opened 2020-07-02
Add: question numbering based on Sphinx section number.

bjones1:
Here's an implementation of my thoughts in RunestoneInteractive/rs#89.

### DIFF
--- a/runestone/common/runestonedirective.py
+++ b/runestone/common/runestonedirective.py
@@ -50,6 +50,11 @@ class RunestoneNode(nodes.Node):
     pass
 
 
+# Provide a class which all Runestone ID nodes will inherit from.
+class RunestoneIdNode(RunestoneNode):
+    pass
+
+
 # Notes
 # env = self.state.document.settings.env
 # env.config.html_context['course_id']
@@ -186,8 +191,12 @@ application.Sphinx.add_autoversioned_stylesheet = _add_autoversioned_stylesheet
 
 
 def setup(app):
+    # Avoid a circular import. Ick.
+    from .question_number import _insert_qnum
+
     # See http://www.sphinx-doc.org/en/stable/extdev/appapi.html#event-env-purge-doc.
     app.connect("env-purge-doc", _purge_runestone_data)
+    app.connect("doctree-resolved", _insert_qnum)
     app.add_role("skipreading", SkipReading)
     # See http://www.sphinx-doc.org/en/stable/extdev/appapi.html#sphinx.application.Sphinx.add_config_value.
     app.add_config_value("runestone_server_side_grading", False, "env")
@@ -403,3 +412,12 @@ def SkipReading(
     inliner.document.settings.env.skipreading.add(docname)
 
     return ([], [])
+
+
+# Return the Sphinx environment given a node object.
+def env_from_node(node):
+    # Ascend the node tree until we find a node with a ``document``.
+    node_with_document = node
+    while not node_with_document.document:
+        node_with_document = node_with_document.parent
+    return node_with_document.document.settings.env
--- a/runestone/blockly/blockly.py
+++ b/runestone/blockly/blockly.py
@@ -18,7 +18,7 @@ __author__ = "bmiller"
 
 import os
 from docutils import nodes
-from runestone.common import RunestoneIdDirective, RunestoneNode
+from runestone.common import RunestoneIdDirective, RunestoneIdNode
 
 
 # try:
@@ -40,7 +40,7 @@ def setup(app):
 
 
 #'
-class BlocklyNode(nodes.General, nodes.Element, RunestoneNode):
+class BlocklyNode(nodes.General, nodes.Element, RunestoneIdNode):
     def __init__(self, content, **kwargs):
         """
 
@@ -49,7 +49,7 @@ class BlocklyNode(nodes.General, nodes.Element, RunestoneNode):
         - `content`:
         """
         super(BlocklyNode, self).__init__(**kwargs)
-        self.ac_components = content
+        self.runestone_options = content
 
 
 START = """
@@ -155,9 +155,9 @@ END = """
 # in html, self is sphinx.writers.html.SmartyPantsHTMLTranslator
 # The node that is passed as a parameter is an instance of our node class.
 def visit_block_node(self, node):
-    res = START % (node.ac_components)
+    res = START % (node.runestone_options)
     res += CTRL_START
-    for ctrl in node.ac_components["controls"]:
+    for ctrl in node.runestone_options["controls"]:
         if ctrl == "variables":
             res += '<category name="Variables" custom="VARIABLE"></category>'
         elif ctrl == "":
@@ -169,11 +169,11 @@ def visit_block_node(self, node):
         else:
             res += '<block type="%s"></block>\n' % (ctrl)
     res += CTRL_END
-    res += END % (node.ac_components)
+    res += END % (node.runestone_options)
     path = os.path.join(
-        node.ac_components["blocklyHomePrefix"],
+        node.runestone_options["blocklyHomePrefix"],
         "_static",
-        node.ac_components["divid"] + ".html",
+        node.runestone_options["divid"] + ".html",
     )
     final = (
         '<iframe class="blk-iframe" seamless src="%s" width="600" '
--- a/runestone/chapterdb/dbchapterinfo.py
+++ b/runestone/chapterdb/dbchapterinfo.py
@@ -161,7 +161,12 @@ def env_updated(app, env):
         doctree = env.get_doctree(docname)
         for section in doctree.traverse(docutils.nodes.section):
             updated_docs.append(docname)
-            title = section.next_node(docutils.nodes.Titular)
+            # Find the section number of the current document. See `../common/question_number.py` for details.
+            secnum_tuple = env.toc_secnumbers.get(docname, {}).get("")
+            # Gievn a section number as a tuple, such as ``(1, 2, 3)``, turn this into the string "1.2.3 ".
+            secnum_str = ".".join(map(str, secnum_tuple)) + " " if secnum_tuple else ""
+            # Prepend it to the title.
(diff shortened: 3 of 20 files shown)


runestone/common/runestonedirective.py
```diff
@@ -241,7 +248,7 @@ def getNumber(self):
 
         env.assesscounter += 1
 
-        res = "Q-%d"
+        res = "Q\ :runestone_secnum:`%d`"
```
bnmnetp, line 251: I think this will just end up with the question in the rendered html looking like:

Q :runestone_secnum:3

Unless that :runestone_secnum: gets inserted *before* we recursively process the text of the question. But I think it might be too late as we are just dumping this into options["qnumber"] which gets handled when we make the html.  Or not at all for many directives. 

Also if we go this way then there are lots of changes that will need to be made in dbchapterinfo.py where we populate the chapters and sub_chapters tables.

bjones1: This works fine for mchoice and fitb questions, since both of those run the reST parser on the text inside the question. I don't know about the other question types; I look at Parson's problems, and that dumps the question text directly into HTML (is this ever checked for XSS?)

I agree that the qnumber won't be correct, but it would be possible to fix that up during the doctree-resolve phase.

Mainly, I'm unhappy with the idea of yet another tie to the chapter/subchapter structure, since this diverges from what Sphinx does in terns of numbering and is something we know we'll have to fix at some point. This PR was a stab at getting us one step closer to aligning with Sphinx.

bjones1:
I just revised this. It should now be able to fix section numbering in directives that don't process their content as reST. It also factors out the call to `addQuestionToDB` until the doctree is resolved and question numbers can be correctly inserted.

There was a circular include cycle between `runestone.server` and `runestone.common` that caused a lot of the noise in this approach.

bjones1:
Better now. Things not done that I'd like @bnmnetp to work on:

- [ ] The database needs to be updated with revised question numbers -- right now, my code just prints the revisions, but doesn't do them.
- [ ] Questions which don't parse their content as reST need to in order to get correct question numbers.

bjones1:
I had a few additional refinements to make this a bit nicer, but won't do anything until Monday at the earliest.

1. Make the directive something like 

```
:qnumber_format:`Q{}`
```

The use the `format` command to insert the section number.

2. Instead of `re`, make a node visitor so that all reST is rendered correctly, then insert the HTML.

runestone/common/runestonedirective.py
bnmnetp: Why create a new directive (:runestone_secnum:)? Is this ever something that an author would write?  It seems like we are just adding in this new directive to be processed later, when we could just add the number directly to the question.

I think we are setting ourselves up for more work by conflating the following two goals:

1. Have the labels on questions match the section numbers generated by sphinx in the table of contents, and make those display correctly in the assignment building interface.
2. Process the question description for each question recursively with sphinx.

To be clear I am in favor of both goals.

We (all of us at the hack week at Luther) made the decision to follow this chapter/subchapter structure years ago and it works just fine for those that color inside those lines.  I realize that is not how you want to structure your book, but it doesn't help to complain that something is broken when you are using it in a way that it wasn't intended to be used.

It seems to me that the only way out of the sort of technical debt that we have incurred with our chapter/subchapter decision is to come up with a long term solution and then work incrementally toward that. Making things incrementally better and making compromises for compatibility as we go.  The large user base that I have to support on runestone.academy makes it hard to make wholesale breaking changes that will mess up people in the midst of teaching a course.

bnmnetp:
I don't see any reason to use DAL in the components.  We consistently use SQLAlchemy here and it works well without needing model files or introducing the database folder and all of that baggage.

runestone/common/runestonedirective.py
bjones1: Thanks for the feedback! I always appreciate other perspectives to help refine and improve my ideas.

My goal aligns with your: look for a long-term solution to paying of the chapter/subchapter technical debt. My strategy for accomplishing it one small step at a time is to gradually transition from chapter/subchapter numbering to Sphinx section numbering; if books (unlike mine) follow the Runestone rules, these are identical. Since we need to add code to number questions, this seems like a good point to use Sphinx section numbers as a part of this overall strategy. However, this does means conflating items 1 and 2, which may be too much to bite off at one time. What are your thoughts on the best way to accomplish this long-term solution?

My implementation was based on this strategy: use Sphinx section numbers as the prefix for question numbers. Since Sphinx section numbers aren't available during parsing, I created a role (which would only be used internally) which can be resolved when the section numbers are available. Again, I'm open to other approaches and can abandon this if you see this as moving us away from this long-term goal. I'm also happy to chat a bit more if you'd like to discuss the approach.

In terms of DAL, it seems to me that having developers know just DAL, instead of both SQLAlchemy and DAL, to develop would simplify things -- the components and the server could share code for some of the tables, queries would be similar, etc. For the same reason, it makes sense to me to do a server rewrite using DAL, not SQLAlchemy. However, this is more a long-term suggestion as well -- ways to make the code easier to maintain in the future.

bnmnetp: My initial thought at where to go with this was to use the sphinx section numbers to populate the chapter and subchapter tables with numbers, ensuring a match.  Then to use the qnumber field of the question table to store the label that we want to display.  That may be flawed for sub-subchapters, more investigation is needed.  That label could then be plugged into the question text when we are walking the nodes and know the numbering for everything.

It is important to have the question label information in the database so that we can use it in the assignment builder and the assignment builder matches the book pages and the table of contents.

We also need to think about backward compatibility for some books that right now do make use of the qnum directive to set a prefix/postfix and plug in a formatted label for the question today.  That required a lot of discipline on the part of the author, nevertheless some authors have used that fairly carefully.  So my thinking was that the qnum directive would override the default of using the sphinx section numbers + question number for the default.

We also need to hash out where these numbers should appear.  Should they be at the beginning of the question text?  Personally I like them a lot better where they appear today, which is below the component.  But I'd like @barbarer and @bhoffman0 and @pearcej to weigh in if they have strong opinions about this.

bnmnetp: On thing I was thinking about with respect to sub-sub sections. was whether or not we out to think about the book being composed of chapters and pages instead of chapters and subchapters.

I'm not sure how you have structured your book.  Does a sububchapter always live in its own file?  Or can you have subsubchapters on the same page?  If they are on the same page then we need a way to figure out how to number the questions that respects those subsub boundaries.  We would not want to have

Question 2.3.11.1
Question 2.3.11.2
then
Question 2.3.12.3
- what happened to 2.3.12.1 and 2?

Should we just use the chapter and subchapter as part of the question number ignoring sub subchapters In question numbering?

bnmnetp: With respect to parsing the question text.   I wonder if it would be possible to have every question use the convention in activecode?  Question text followed by a line with `~~~~`. This would make it easier than handling some questions that use `:question:` for the question.

This would let us code this once rather than handling it in each directive.

bjones1: I'm certainly in favor of using Sphinx section numbering wherever it's easy to gather for Runestone. That's certainly a step in the right direction.

What is a question label? Is it a question number, like question 2.3.11.1, or something else? As far as I know, question numbers currently appear at the beginning of a question, while the div_id (label?) appears at the bottom of the question.

My sub-subchapters are always separate files, due to the requirements of the literate programming question type I use.

I think using a consistent method through all the Runestone components, such as dividing question text from other parts with a `~~~~`, is a great idea! However, doesn't it also mean we'll need to update everything question in our books to follow this new format? That seems painful.

bnmnetp: Some authors use the `..qnum::` directive on each page.

```
.. qnum:: 
   :prefix: func-
   :start: 1
```

For this page every question is labelled/numbered func-X

The feedback from the majority of teachers was that they found that confusing.  So I turned it off completely for fopp.

They seem to like the `Activity X:  Multiple Choice (div_id_here)` that appears below the question.  -- Those are numbered using a simple CSS counter on each page.  My assumption is that they would equally like
`Activity 2.3.4: Multiple Choice (div_id_here)`. Especially if 2.3.4 also appeared in the assignment builder.

I agree that any modification to the question format would be painful, unless it was easy to automate updating all of our book source.  which in this case I doubt.   It turns out that for purposes of the description it is easy enough to just grab the content up to the first blank line for each question, or to override that just for the activecode and clickablearea questions.

bjones1: That sounds good to me. I don't use `qnum` and like the idea of auto-numbered questions as you propose.

Do you want me to do anything on this, or would you like to code it up?

bjones1:
I did a bit of final polish and pushed the results. Even if we abandon using the question numbers and instead prepend section numbers to question labels, I think some pieces in here will be helpful: having docutils nodes from all components with an id inhert from a RunestoneIdNode; having all such components using a consistent naming convention for storing the component's options; finding and formatting Sphinx section numbers; walking the doctree only for RunestoneIdNode matches.

bjones1:
One last update -- it was a good chance to learn about the NodeVisitor pattern. This flavor correctly supports numbered sections that appear on a given page, rather than assuming the section number is a per-page constant. It drops modifying the question number (and the associated role), instead providing a place to update the component and database with the correct section number for that point in the document.

bjones1:
OK, this is ready from my side. @bnmnetp, would you:

- [ ] Add a `data-label` kind of element to the HTML each component generates, based on the `node.runestone_options["question_label"]` string
- [ ] Have the JS use this for the question label?

Then updated question labels will be ready (but see the server-side updates linked above).

bjones1:
Even though CI failed, it looks like a random confusion (just Py 3.8 because runestone.js wasn't found), not anything due to the code itself. I've made all the changes we discussed today via Zoom, with one exception: I'm not sure how to put the `data-question_label` in `blockly.py` -- would you help? Also, would you look at components that need visitors and add those, per our discussion?

PR #1026, opened 2020-07-13
Selectquestion

bnmnetp:
This PR will implement a set of directives to support automatic exam generation.

Starting with

```
    .. selectquestion:: [fromid, [fromid]+ ]
       :proficiency: randomly choose a question that tests a particular proficiency
       :basecourse: restrict question choices to the current base course
```

More will be added to allow you to select a group of questions that meet some criteria.

For purposes of discussion I did consider two implementation strategies.

* By the server?
	* Write out a page (as a template) during the build
	* During the build also populate the database with the exam specification
	* The exam controller will use the view for this specific exam
	* if a student has not seen the exam how will be know which questions to select?
	* 
* By the javascript?
	* When rendering a question are we inside an exam?
	* get the question and student for this exam
		* the server can check - if there is no entry for this question then look at the criteria and select one as well as store, then return the question
		* the server should make sure there are no duplicated question ids for this exam

The really big downside of the server side approach which led me to abandon it, is that those pages would need to have their own controller that would have to figure out what to serve.  This would limit the use of the selectquestion directive to a small number of pages where a custom controller was written.


bjones1:
Here's some random, unstructured feedback. I'm happy to chat if that's more helpful.

One use scenario: I put in 3 selectquestions for topic A, 2 for topic B, etc; therefore, I want to specify problems by topic, not by divid. Therefore, I see this as a two-part problem:

1. Provide a way to categorize problems / group them into topics. A specific page in a book defines a topic.
    - Each book's problems should be accessible base_course + div_id specifier.
    - In addition, have "books" that are pure, organized problems.
2. Provide a way to pull a problem from a given topic.

The suggested syntax:

This directive will place the specified question(s) on the page in which the directive appears. These questions will therefore be considered by a `select_question` directive (along with all other questions on this page).
```
.. include_question::
    base_course1 div_id1a div_id1b ...
    base_course2 div_id2a div_id2b ...
    ...
```

To select a question from one or more "topics", defined as specific page(s) of book(s):
```
.. select_question:: div_id
    [/base_course1/]path/to/book/page1
    [/base_course2/]path/to/book/page2
    ...
```
where:
- `div_id` is the div_id of the resulting selected question.
- An absolute path indicates that the first element of the path if the base course. A relative path may only refer to questions within its own book.

PR #1027, opened 2020-07-15
Fix the unpairing distractors problem

Coldstream-Louis:
In the current version, if the user solves the problem for the first time. When he or she clicks the reset button, the distractors will be unpaired and all the blocks will be shuffled. The old version could not shuffle the distractors randomly.


bnmnetp:
@ericsonga -- Are you good to go with this??

Coldstream-Louis:
This commit has a problem of displaying the decoration of the paired distractor incorrectly. I have made a change to solve this problem and will pull a new request.

PR #1040, opened 2020-08-05
Unpaired distractors & Area width adapting

Coldstream-Louis:
1. Fixed unpaired distractors problem. Now it will randomly shuffle all the blocks
2. Fixed the Area width adapting. The width of the source/answer areas will be adjusted by the length of the code line inside each block.


Coldstream-Louis:
Don't know why this is still failed

Coldstream-Louis:
I also have failed the test dnd1 of drag and drop test when I run "pytest" locally, but here it shows the error is in dnd2

PR #1050, opened 2020-08-21
Test Case for Timed File in Runestone Directory

willxli:
- Created test case to test for conditions of 1 question timed exam. (Made sure there was 'start' button, user was able to click on MC question, submit answer by pressing 'Finish', alert pops up to make sure user wants to submit and make sure there is feedback on MC question)


bnmnetp:
Thanks for the PR, always nice to have more tests.  This will be great to build on.

PR #1077, opened 2020-09-24
Parsons height width fix to handle different computers rendering different typefaces

vqum:
Previously manual calculation of Parsons heights and widths was built to only address one typeface, but different computers may render different typefaces along the priority list of `Menlo, Monaco, Consolas, "Courier New", monospace;` as set in bootstrap.css. Additional code has been added to determine which typeface will be rendered by the computer and then set that typeface to be used for calculating Parsons heights and widths.

### DIFF
--- a/runestone/parsons/js/parsons.js
+++ b/runestone/parsons/js/parsons.js
@@ -1497,7 +1497,7 @@ export default class Parsons extends RunestoneBase {
                 var fontFamily = linesItem[linesIndex].children[0].fontFamily;
                 var tempCanvas = document.createElement("canvas");
                 var tempCanvasCtx = tempCanvas.getContext("2d");
-                tempCanvasCtx.font = fontSize + fontFamily;
+                tempCanvasCtx.font = fontSize + " " + fontFamily;
 
                 // Increment Parsons area height based on number of lines of text in the current Parsons block - Vincent Qiu (September 2020)
                 var singleHeight = 40;
--- a/runestone/parsons/js/parsons.js
+++ b/runestone/parsons/js/parsons.js
@@ -251,9 +251,29 @@ class ParsonsLine {
         // Pass this information on to be used in class Parsons function initializeAreas
         // to manually determine appropriate widths - Vincent Qiu (September 2020)
         this.view.fontSize = window.getComputedStyle(this.view, null).getPropertyValue('font-size');
-        this.view.fontFamily = window.getComputedStyle(this.view, null).getPropertyValue('font-family');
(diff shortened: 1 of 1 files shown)


runestone/parsons/js/parsons.js
```diff
@@ -1472,9 +1492,9 @@ export default class Parsons extends RunestoneBase {
...
                 }
 
-                // Create a canvas and set the passed through fontSize and fontFamily in order to later measure text width - Vincent Qiu (September 2020)
+                // Create a canvas and set the passed through fontSize and fontFamily in order to measure text width - Vincent Qiu (September 2020)
                 var fontSize = linesItem[linesIndex].children[0].fontSize;
-                var fontFamily = linesItem[linesIndex].children[0].fontFamily.split(",")[2];
+                var fontFamily = linesItem[linesIndex].children[0].fontFamily;
                 var tempCanvas = document.createElement("canvas");
                 var tempCanvasCtx = tempCanvas.getContext("2d");
                 tempCanvasCtx.font = fontSize + fontFamily;
```
bnmnetp, line 1500: There needs to be a space between the fontsize and the fontFamily if they are jammed together it doesn't recognize them and the sizes are still wrong.

vqum: Oops, I did catch that in my testing code but forgot to copy that change into the commit.

bnmnetp:
This looks like it works for me.  Do you think it is as reliable as the solution in https://stackoverflow.com/questions/3368837/list-every-font-a-users-browser-can-display. ??

vqum:
Is there a particular solution in that thread that you are referring to?

1. Regarding 1st response submitted by alex of using Flash with JavaScript, I heard Flash will no longer be supported after 2020, so i don't think this wouldn't be sustainable. Unless the Flash being used in that solution refers to something else that I'm not familiar with.

2. This response submitted by Marko of comparing text widths and heights is essentially what my solution is doing, and I saw many examples across the web of doing some form of comparing widths and heights in order to match typefaces. That said, my solution does omit comparing heights and comparing to monospace and sans-serif because I thought it wouldn't be necessary based on my understanding that the list of possible rendered typefaces is short (the 6 of Menlo, Monaco, Consolas, Courier New, monospace, and browser default). I also encountered an issue when initially I did include comparing to monospace where the width of monospace matched the width of Consolas resulting in the code assuming Consolas was not the correct typeface. If this is the solution that you are referring to, then I could take another look at this to see whether my solution could be made more reliable.

3. This response submitted by Youssef AbouEgla of using document.fonts is not Internet Explorer compatible.

I only read that far, but please let me know your thoughts.

PR #1074, opened 2020-09-11
Bug Fixes - Parsons problems heights, widths, and image displays (Vincent Qiu September 2020)

vqum:
Many Parsons problem areas were displaying with significantly more height than necessary taking up more screen space and with less width than necessary causing lines of text to overflow. Now calculations for Parsons problem areas heights and widths have been reprogrammed within parsons.js to resolve this issue.

Some Parsons problems include images (e.g. https://runestone.academy/runestone/books/published/StudentCSP/CSPRepeatTurtles/patterns.html) that cause misalignment of the problem text and Parsons problem areas. Now images are being checked for and , if found, are being moved to their own <div> within parsons.js, and the new <div> has a corresponding class within parsons.css to resolve this issue.

additional commit on September 18, 2020:
Incorporated William Li's changes to parsons.js from previous pull request: https://github.com/RunestoneInteractive/RunestoneComponents/pull/1051

PR #1100, opened 2020-11-13
Parsons height display bug fix (Vincent Qiu November 2020)

vqum:
An edge case of Parsons display that was missed in previous changes https://github.com/RunestoneInteractive/RunestoneComponents/pull/1074 and https://github.com/RunestoneInteractive/RunestoneComponents/pull/1077 is when a Parsons line is just barely longer than the set width limit. Lines longer than width limit are normally overflowed to next line and the Parsons background area normally grows to accommodate the increased height of the blocks, but this was not happening with lines just barely longer than the width limit.

On MacOS, an example was discovered in Python for Everybody > 15 Databases > 15.15 Mixed-Up Code Questions > Q-2 and Q-4 (https://runestone.academy/runestone/books/published/py4e-int/database/mixedupCode.html) where the lines similar to `cur.execute('INSERT INTO Dogs (name, breed) VALUES (?, ?)',` overflow the last few characters to the next line, causing an increase the height of the entire Parsons problem, but the background Parsons area does not grow to accommodate it.


vqum:
Could the checks be ran again? It looks like the failed build is similar to a previous pull request https://github.com/RunestoneInteractive/RunestoneComponents/pull/1074 that I submitted where it failed at something unrelated to the change being submitted.

This build passed for Python 3.8, but failed for Python 3.7:
https://github.com/RunestoneInteractive/RunestoneComponents/runs/1397691967
image

These errors occurred at four DragAndDropQuestion_Tests:
https://travis-ci.com/github/RunestoneInteractive/RunestoneComponents/jobs/439524428
image

All passed when I ran `pytest` locally just now in `runestone\parsons\test`:
image

------------------------------------------------

Alternatively, the commit that failed the check was a change that I had previously made in `runestone\timed\js\timed.js` that I did not intend to include in this pull request. Would there have been a better way for me to remove this change other than to undo the code changes and then push that undone file into my repository? Could it have been this that caused the check to fail?

PR #1104, opened 2020-12-03
Fix: Only restart question numbers on numbered sections.

bjones1:


bnmnetp:
What are the implications of this PR?  Is it a performance thing?  Will this change the numbering on any of the existing books?  Under this How will exercises be numbered in sections that are not numbered?

bjones1:
It's correctness, not performance. Here's the incorrect case:

1.1 A heading
   Question 1.1.1
   Question 1.1.2

An unnumbered subheading
    Question 1.1.1 (incorrect -- the old code restarted question numbers at EVERY heading, numbered or not)

With this fix, the last question will now be 1.1.3.

bnmnetp:
Thanks!   I know its a bit lazy as I could figure it out with some work,  but just having a high level explanation really helps!

bjones1:
My pleasure! I also appreciate your review -- knowing there's another set of eyes on my code is reassuring.

PR #1105, opened 2020-12-03
Fix: Return data from the server, instead of the HTTP response, from …

bjones1:
…logBookEvent.


bjones1:
This actually passed -- the failure was a hiccup on one of the tests ([missing runestone.js](https://travis-ci.com/github/RunestoneInteractive/RunestoneComponents/jobs/454036828#L627)). @bnmnetp, would you mind restarting the failing test?

runestone/fitb/js/fitb.js
```diff
@@ -241,10 +241,9 @@ export default class FITB extends RunestoneBase {
         }
         if (logFlag) {
             // Sometimes we don't want to log the answer--for example, when timed exam questions are re-loaded
-            let ret = await this.logCurrentAnswer();
+            let data = await this.logCurrentAnswer();
```
bnmnetp, line 244: The first promise resolves to a response object.  You need to await the call to response.json() to get the json object.  I'm not following why you made this change.

bnmnetp: I guess you moved that to logBookEvent?

bjones1: What we want from logBookEvent is the returned JSON -- it's already checking the response, so I see no need to return the response itself.

Do you want me to look at updating logRunEvent?

bnmnetp: That would be great.  I definitely think those two should be consistent in what they return.

bjones1: Looking through the code, most calls of `logRunEvent` aren't awaited, meaning the calls aren't guaranteed to complete. Should all these be made using either `await` or `.then`, or do I misunderstand the code?

bjones1: Also, I wanted to do a better job of documenting this new interface / approach. Do you have anything written up on this, or should I create it?

bnmnetp: For better or for worse, the vast majority of the time we have never cared or checked the results of logging.  Just fire off the logging event and move on without worrying if it was successful or not.

bnmnetp: I have not written anything up, other than the additional comments I may have added here and there as I was working on it.

bjones1: OK, sounds good. I'll run with that. Thanks!

bjones1:
How does this look?

runestone/mchoice/test/test_assess.py
```diff
@@ -63,12 +62,6 @@ class MultipleChoiceQuestion_Tests(RunestoneTestCase):
     def test_ma1(self):
         """Multiple Answer: Nothing selected, Check button clicked"""
         self.driver.get(self.host + "/index.html")
-        wait = WebDriverWait(self.driver, 10)
```
bnmnetp, line 66: I'm curious as to why you removed this block.  Our tests are riddled with the kinds of timing problems that this is supposed to fix.

bjones1: I added a [default implicit wait of 5 seconds](https://github.com/RunestoneInteractive/RunestoneComponents/pull/1105/files#diff-80bc6a407ddd8b12c6f879353243ef7e16fd852f16d0ad34a06975d28f1272f6R230), which should be fine for most cases. Therefore, we can remove most of the explicit waits, which are rather ugly code-wise.

bnmnetp: I think that works for some things but not everything.  Particularly when we are getting the text from an element and then using `self.assertIn("foo", element.text)`  See https://github.com/RunestoneInteractive/RunestoneComponents/commit/4074b39202123749462f380c9c0bc71edbeb83be for my updates to the activecode tests that should fix that in a foolproof way.

bjones1: I agree -- more specific conditions can/should be added.  I think with the implicit 5 second wait, you just need to assert that the expected condition is met, so:

```
       wait = WebDriverWait(self.driver, 10)
        wait.until(
            EC.text_to_be_present_in_element((By.ID, "test1_stdout"), "Hello")
        )
```

can (I think) become

`assert(EC.text_to_be_present_in_element((By.ID, "test1_stdout"), "Hello"))`

per the docs on [explicit waits](https://selenium-python.readthedocs.io/waits.html#explicit-waits).

bnmnetp: That does not seem to work for me.  If I change hello to "sdlkfjsh" the assert still works.  even though the only text in the element is "Hello World". Whereas with the wait.until if I look for the "sdlkfjsh" it times out and fails.

bnmnetp: A slightly cleaner solution for the text fields is to create a `self.wait` object in setUp.  Then we can do:

```
        self.wait.until(
            EC.text_to_be_present_in_element((By.ID, "test1_stdout"), "Hello World"),
            message="Did not find expected text",
        )
```
If we are only looking for one thing in the output then there is no need for a separate assert as it will cause the test to error if the text is not there after 5 seconds.

If you want to look for multiple things in the output then you can wait for the first one and use simple asserts for the rest.

bjones1: OK, thanks for trying. I like your idea of creating a `wait` object in `setUp` -- that seems like a good way to go.

bnmnetp:
I must have caused this conflict, but it is really stupid.

PR #1107, opened 2020-12-07
Docs

bjones1:
I'm working on improving documentation, and will submit periodic PRs. Your comments and feedback on how these docs come together would be very helpful!


bjones1:
Note that CI failed, even though the changes are to comments only -- looks like more timing problems in CI.

bnmnetp:
I think we need to commit to fixing these as we encounter them.  I am getting really frustrated and tired of having to re-start builds all of the time, and I can't just fix everything.

This weekend it took me six tries to get both 3.7 and 3.8 to pass all of the tests, each time with a different random test failing.  I feel like the little Dutch boy trying to plug all the holes in the dike.  There are good solutions to these timing problems but we need to take the time to fix them as they pop up.

bjones1:
That makes sense. Random CI fails are extremely frustrating. I would suggest adding the implicit wait first to see if that helps. Looking at the logs, I'm seeing two classes of failures:

1. Complaints that runestone.js isn't present. I suspect it's something else, but it appears to be something that's not Selenium timing related
2. Timing problems in Selenium.

No answers yet, but I'll start looking at this more carefully.

bnmnetp:
Can you point me to one of the runestone.js errors?  I am pretty sure that is an indication that `npm run build` either failed or was not run for some reason.  I added an error message for a missing runestone.js just for this case.

bjones1:
https://travis-ci.com/github/RunestoneInteractive/RunestoneComponents/jobs/455457111#L632

bnmnetp:
Weird, that error message comes from line 154 in `__main__.py`. but in the log you sent me I can see that runestone.js was built successfully right before.

bjones1:
Exactly, making me think that something else is going on. But I think it's a separate type of error than the Selenium timeouts.

bnmnetp:
I think it is in `unittest_base.py` in setUpModule we do `npm run build` and immediately thereafter `runestone build --all`. But Popen is non-blocking!

Should we change those two calls to use subprocess.call instead?  That will wait until the commands have completed?

Separately:
It is dumb to have that extra `npm run build` in the travis.yml file.   I think having it in the setup is much better.  I'm sure I am the one that added it to travis as well.

bjones1:
Nice catch! Yes, let's keep setup in the unit test, so that local testing will always use current JS and remove it from Travis. We should definitely change it to a more modern call -- I'd suggest `subprocess.run`.

bjones1:
I see a failure at https://travis-ci.com/github/RunestoneInteractive/RunestoneComponents/jobs/455791475#L617. Before I submit a PR to fix  this, would you apply the fixes (subprocess.run, provide a pre-built wait in setUp) we've discussed earlier?

bnmnetp:
just pushed it.

PR #1108, opened 2020-12-07
Selenium fixes

bjones1:
This prints out npm run build output, in addition to a bit of updating.

### DIFF
--- a/runestone/mchoice/test/test_assess.py
+++ b/runestone/mchoice/test/test_assess.py
@@ -1,15 +1,36 @@
-"""
-Test Multiple Choice question directive
-"""
+# ****************************************************
+# |docname| - Test Multiple Choice question directive
+# ****************************************************
 
 __author__ = "yasinovskyy"
 
 from unittest import TestCase
-from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.common.by import By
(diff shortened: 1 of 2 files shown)


bjones1:
The tests now pass, which *may* mean one of the Selenium timing problems are fixed.

runestone/mchoice/test/test_assess.py
```diff
@@ -1,16 +1,37 @@
...
 
 from unittest import TestCase
-from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.common.by import By
 from runestone.unittest_base import module_fixture_maker, RunestoneTestCase
 
 mf, setUpModule, tearDownModule = module_fixture_maker(__file__, True)
 
+
+class element_has_css_class:
```
bnmnetp, line 14: Ooooh nice!

bnmnetp: Actually this seems like it would be good to have in unittest_base so that we can use it in other tests

bjones1: Good point. I'll move it there.

bjones1:
Sigh, a test failure popped up and even the 5 second wait didn't help. I'll test locally to see if I can reproduce this.

Issue #1113, opened 2020-12-17
pip install fails on Python 3.9.x on Windows 10


schellenberg:
Currently, you cannot install runestone on Windows 10 with Python 3.9. This is due to the version of lxml being used (4.5.1). However, the current version of lxml (4.6.2) can be installed via pip just fine. 

Can you update the requirements.txt file to either:
- lxml==4.6.2               # assuming this version doesn't have any backwards-incompatibilities

OR  

- lxml>=4.5.1

Thanks!

bnmnetp:
Or you could make a PR? 😄 Its hard for one person to keep up.

PR #1114, opened 2020-12-17
Update requirements.txt

schellenberg:
Update lxml version to allow for use in Python 3.9 on Windows 10

https://github.com/RunestoneInteractive/RunestoneComponents/issues/1113#issue-770308169

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ docutils==0.16            # via codechat, sphinx
 idna==2.9                 # via requests
 imagesize==1.2.0          # via sphinx
 jinja2==2.11.2            # via sphinx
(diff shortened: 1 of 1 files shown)


bnmnetp:
requirements.txt and requirements-dev.txt are generated by pip-compile from requirements.in and requirements-dev.in.

Running `pip-compile --upgrade requirements.in requirements-dev.in --output-file requirements-dev.txt` and `pip-compile --upgrade requirements.in  --output-file requirements.txt` will generate and attempt to upgrade dependencies if possible.  Since this is a dependency of CodeChat and not one that we specify directly that is the safe way to do it.

bnmnetp:
Done

PR #1136, opened 2021-02-04
Cellbotics

bjones1:
This JavaScript plus one directive adds support for reading sensors on smart devices and invoking functions on an Arduino connected via BLE. This is the built version of https://github.com/RunestoneInteractive/skulpt/pull/5.


bjones1:
I updated this, since it was out of date. With webpack, this JS will only be loaded when a page refers to it; @bnmnetp, would you consider merging it, since there's little impact?

bjones1:
Thanks!

PR #1148, opened 2021-03-15
Improve Runestone testing

bjones1:
This improves timed exam unit tests, as a first step toward #1146. Next up will be `selectquestion` testing.

Add: more testing for a timed exam.
Fix: Improve unit testing approach:
       Run webpack before each test
       Properly close the Selenium driver
       Provide a wait_until_ready method


bjones1:
@bnmnetp, it looks like the unit tests are confused here too -- Py 3.7 passes, Python 3.8 doesn't. We're definitely in need of some unit test improvements.

bjones1:
FYI, this passed on Travis (https://travis-ci.org/github/bjones1/RunestoneComponents/builds/763045835).

bjones1:
Yikes. Apparently, a delay makes the tests pass...

...which is why we need a better test methodology. The new code allows components to indicate when they're ready. I didn't include it in this fix, since a shortanswer question is used by selectquestion and is a big, potentially breaking change. When this is merged, I'll submit a PR to update the poll component to use this "smart" approach. I chose pool because the poll component is a low-risk place to start.

.travis.yml
```diff
@@ -19,7 +19,6 @@ script:
 - "python -m pip install -U pip"
 - "pip install -e ."
 - "pip install -U -r requirements-dev.txt"
-- "npm run build"
```
bnmnetp, line 22: This file could be deleted since I moved us over to using github actions.

PR #1150, opened 2021-03-17
Improved testing 2

bjones1:
This PR presents the new test environment, enabling Runestone Components and the server to share Selenium test code. It also updates the poll component to use this new framework.


bjones1:
When this is merged, I'll submit a PR for the server-side test of this question.

runestone/common/js/runestonebase.js
```diff
@@ -212,6 +219,18 @@ export default class RunestoneBase {
             this.checkLocalStorage(); // just go right to local storage
             this.csresolver("local");
```
bnmnetp, line 220: Now that I look at this more carefully, I *think*  that the call to `this.csresolver()` is solving the same problem as adding the runestone-component-ready class and triggering: `runestone:component-ready`.   The difference is that it is resolving a promise that other objects could `await` or `.then`

*Maybe* using that mechanism is better than adding another one?  Note that the `csresolver` is pretty new and something I added recently which does not look widely used....

runestone/poll/js/poll.js
```diff
@@ -59,7 +60,7 @@ export default class Poll extends RunestoneBase {
         this.containerDiv = document.createElement("div");
         this.pollForm = document.createElement("form");
         this.resultsDiv = document.createElement("div");
-        this.containerDiv.id = this.divid + "_container";
+        this.containerDiv.id = this.divid;
```
bnmnetp, line 63: I think this is how it should have been all along.

runestone/poll/js/poll.js
```diff
@@ -212,6 +213,7 @@ export default class Poll extends RunestoneBase {
             }
             $(this.resultsDiv).append(list);
         }
+        this.indicate_component_ready();
```
bnmnetp, line 216: If its called in the base class why call it again here??  Shouldn't there be one and only one indication of when the component is ready.  Otherwise we may get multiple firings of things that should not be done multiple times.

runestone/common/js/runestonebase.js
bjones1: Hmm, interesting. I remember seeing that in the code, but not looking at it very deeply. My thoughts:

- This is really nice. I like the idea of allowing code to `await` on a `csresolver` promise rather than the old jQuery emit logic. Are there any components which don't invoke `checkServer` in the constructor? If so, I'd suggest creating the promise in the constructor, so it can be waited on at any point.
- From a test perspective, there's no clean way to await a JavaScript promise from Python test code. There's probably a hack (set a global variable in JS then poll for it from Python/Selenium), but since there's already good support for waiting for a class to be present, I see no advantage in doing this. So, I'd suggest keeping the code which adds the `runestone-component-ready` class. In the future, I could see adding a CSS style to visually indicate when a component is fully loaded, making this useful for more than just testing.
- This promise doesn't always tell the whole story -- for example, the poll component does some additional server-side operations (loading from the `getpollresults` endpoint), so just awaiting on `csresolver` doesn't work for all components.

runestone/poll/js/poll.js
bjones1: I agree! My goal was to make only the changes necessary / reasonable to the components (provide a single, unique ID, etc.).

runestone/poll/js/poll.js
bjones1: It's only called in the base class if `this.checkServer(event_name, true)` is invoked (note the `true` parameter, which defaults to `false` if not provided). This component doesn't call it, instead calling `checkLocalStorage()`.

Here's the flow of this component, as I understand it:

- Constructor called
- Constructor invokes `checkPollStorage()` (line 29)
- In `checkPollStorage()` (line 219):
    - If there's nothing in local storage, indicate the component is ready (line 234).
    - Otherwise, load from the `getpollresults` endpoint (line 228)
        - On success, invoke `showPollResults()` (line 157); at the end of this function, indicate the component is ready (line 216)
        - On failure, indicate the component is ready (line 232).

...which is one of the more complex flows I've seen in the components.

runestone/common/js/runestonebase.js
bjones1: I just pushed 646d3611 to address this. It uses a different promise (`this.component_ready_promise`), so that we can distinguish between `checkServer` competing (with the `this.checkServerComplete` promise) and the component being ready (which differs from `checkServer` completion for some components, such as poll).

bjones1:
Note that the test failure is again due to timing problems with tests. The shortanswer question type doesn't (yet) have a way to wait until the component is ready to begin testing. I will increase the delay on the shortanswer test to see if that passes.

runestone/common/js/runestonebase.js
bnmnetp: checkServer is not used by activecode or any of the non-gradable components.  So I think having a separate promise is good, and all components can then use their own logic to ensure that when the component is ready then the component_ready_promise can be resolved.  For those that do use checkServer we should ensure that checkServer is resolved before resolving component_ready.

bnmnetp: I agree that we should not use crazy hacks for testing when there is a tried and true method for us to use.  Adding a class is fine and as you say may have good benefits later.

bjones1:
...and the test passes now. Manual timing tweaks are so painful.

runestone/common/js/runestonebase.js
bjones1: Sounds good. Since `checkServer()` calls `indicate_component_ready` only after the server data is loaded, we're in good shape!

bnmnetp:
An idea for a minor tweak -- why not have component_ready_resolve return either the div_id of the component, or `this` the component itself?

bjones1:
Reasonable, but doesn't the caller by definition have `this`  in order to invoke `await this.component_ready_promise`? Given `this`, it's easy to access `this.componentDiv` or `this.componentDiv.id`.

bnmnetp:
Yes, I was just about to delete that comment when you replied.

PR #1162, opened 2021-03-27
Improved testing 4

bjones1:
Continuing #1148.

Add: smart waits
Port: to pytest.


bjones1:
Does this need any server-side testing? It looks like everything here is client-side.

bnmnetp:
Yes, this is just for presentation purposes.

PR #1163, opened 2021-03-28
Improved testing 5

bjones1:
Another from #1148. This required a rename of one member variable. Other than that, it's pretty simple.

Add: Smart waits
Port: To pytest.


bjones1:
Like #1162, I assume this needs no server-side tests, since it's client-only code -- please confirm.

bnmnetp:
correct.

PR #1164, opened 2021-03-28
Improved testing 6

bjones1:
Following #1148.

This component needed a few more fixes than previous ones:
- Remove a duplicate ID
- Have it inherit from Runestone Components
- Switch from the containing div as a jQuery object to a standard DOM reference (this was the only component not using a DOM reference to the containing div).

Add: Smart waits
Port: To pytest

### DIFF
--- a/runestone/showeval/README.md
+++ b/runestone/showeval/README.md
@@ -1,7 +1,7 @@
 <h2>ShowEval</h2>
 
 ```html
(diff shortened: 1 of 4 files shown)


bjones1:
Again, I'm assuming that no server-side testing is needed, since I see no server interaction.

PR #1170, opened 2021-04-08
Clean: DRY tests.

bjones1:
Some clean-up of the testing framework.


runestone/poll/test/test_poll.py
```diff
@@ -1,25 +1,20 @@
-def _test_poll(selenium_utils, id):
+def test_poll(selenium_utils_get):
```
bnmnetp, line 1: The server side tests are still trying to import `_test_poll`. Not sure which way you want to go here

bjones1: This is fixed in https://github.com/RunestoneInteractive/RunestoneServer/pull/1634.

PR #1174, opened 2021-04-12
Improved testing 13

bjones1:
Add: Better testability for drag-n-drop components.

Another in #1148.

### DIFF
--- a/runestone/dragndrop/README.md
+++ b/runestone/dragndrop/README.md
@@ -2,17 +2,17 @@
 
 ```html
 <ul data-component="dragndrop" id="dd1">
-    <span data-component="question">The Question goes here.</span>
-	<span data-component="feedback">This is feedback that is displayed when this is answered incorrectly.</span>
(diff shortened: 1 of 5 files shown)


runestone/dragndrop/README.md
```diff
@@ -2,17 +2,17 @@
 
 ```html
 <ul data-component="dragndrop" id="dd1">
-    <span data-component="question">The Question goes here.</span>
```
bnmnetp, line 5: Ooooh.. Thanks for this... I have a feeling this fixes another seldom seen but nasty bug that I've encountered a few times with timed selectquestions....  We'll see.

Its great to have another set of eyes looking at this stuff... some of which hasn't been touched for ages!

bnmnetp (approved):

runestone/dragndrop/README.md
bjones1: My pleasure! I hope this helps. I have more tests working for the timed component as well, so that will have better test coverage.

PR #1175, opened 2021-04-13
Improve testing 14

bjones1:
Updates for activecode -- mostly, tweaks to imports and a rename of one div. Another from #1148.


runestone/activecode/test/test_activecode.py
```diff
@@ -1,242 +1,209 @@
...
-            # """,
-            #     interceptor,
-            # )
-            self.driver.execute_script("window.scrollTo(0, 0);")
-            rb.click()
-
-        ta = t1.find_element_by_class_name("cm-s-default")
-        self.assertIsNotNone(ta)
-        self.driver.execute_script(
-            """window.edList['test1'].editor.setValue("print('GoodBye')")"""
```
bjones1, line 99: @bnmnetp, I'm having trouble with this line when doing server-side testing with selectquestion. While the component runs as expected in a selectquestion, this statement doesn't update the text in the activecode window; it works fine here, though. Any ideas on another way to programmatically change the text in an activecode?

bnmnetp: Hmmm, What is the problem?  Is there a javascript error?

Could we use a method similar to what you use for fill in the blank?

bnmnetp: One way that should definitely work would be to send a `runlog` entry to the server to pre load history and then click on the load history button for the activecode.

But it is still not at all clear why the execute_script method is not working.

bjones1: The tricky part is getting text to CodeMirror. It  does lots of JS tricks that make the actual element that receives keystrokes hidden behind a bunch of divs that display line numbers, syntax highlighting, etc. So, the Selenium `send_keys` method doesn't work (to the best of my knowledge) since we can't find the correct HTML element to send it to.

In the client-side tests, the JS here works just fine. When run outside a selectquestion, it's also fine. But, evidently the way selectquestion loads the activecode makes this not work.

I was hoping you'd have some clue as to why -- this is pretty tied up inside selectquestion/activecode, and is pretty far from the parts of the code I know...

bjones1: Hmm, the `runlog` approach seems reasonable. It's fairly easy to debug -- just insert an `import pdb; pdb.set_trace()` just before the failure, then open the Chrome browser, open the debug pane, and play. Since the server is up and running, you can both do client-side playing and server-side playing.

...but, as I said before, I don't really know the system well enough to debug it...

bnmnetp: Having said that it does not look like it is failing?  

In any case I'm assuming you mean on the line right before execute_script?

bjones1: The tests fail only on the server -- I'll push the server tests. (They'll fail, since these changes aren't merged). These tests pass, but will probably need a bit of tweaking at this line to get the server-side tests fully working.

bjones1: Yes, exactly. Do you want to chat briefly via Zoom? I'm on Slack.

bjones1: Note that none of the tests currently fail, since I marked the failing server-side selectquestion test with a skip.

bjones1:
@bnmnetp, would you check? All the tested pass, but the build was for some reason canceled just after they passed. Perhaps re-run the tests?

bjones1:
@bnmnetp, would you take a look when you have time? I can't get this to fail locally, and debugging via GH Actions is quite slow and painful.

bnmnetp:
I've tried re-running multiple times... trying again now.

bjones1:
@bnmnetp, this is finally working. However, a bunch of commits are a mess. After you review it, would you like me to squash and force-push something cleaner?

bnmnetp:
@bjones1 -- something is still not quite right with this.... I just pulled this to my dev environment and used it to build the overview book.  In the timed exam on the Dynamic Questions page (uses selecquestion) I cannot edit or do anything with the 4th question (an activecode) I can't edit it and the run button does not seem to be responsive as well as the Caption is showing which should not be the case in a timed exam.

bjones1:
OK, sounds good. Thanks for doing the manual testing to verify!

I'm getting close to being able to merge a timed question that exercises more of the question types inside it, which we should also be able to run server side as well.

PR #1176, opened 2021-04-14
Improved testing 15

bjones1:
Fix: Make CodeLens more testable. Another of #1148.


bjones1:
Does this need to be tested server-side?

bjones1:
@bnmnetp, I think/hope it was a testing hiccup. I re-ran the tests and all is well. I couldn't reproduce the failures locally either.

PR #1185, opened 2021-04-24
Clean: Move imports to webpack

bjones1:
This is another PR moving toward webpack by first cleaning up some JS imports. The goal: move all imports possible to JS, instead of Sphinx `add_javascript` or `add_css_file` methods. This is a first step; eventually, I'd like everything to be in JS includes.

There's also a fix for one component that wasn't labeled with a `data-component` attribute.

### DIFF
--- a/runestone/selectquestion/js/selectone.js
+++ b/runestone/selectquestion/js/selectone.js
@@ -104,7 +104,7 @@ export default class SelectOne extends RunestoneBase {
         if (opts.timed) {
             // timed components are not rendered immediately, only when the student
             // starts the assessment and visits this particular entry.
-            res = await createTimedComponent(htmlsrc, {
+            res = createTimedComponent(htmlsrc, {
                 timed: true,
                 selector_id: selectorId,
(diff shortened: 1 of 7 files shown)


runestone/selectquestion/js/selectone.js
```diff
@@ -6,8 +6,8 @@
 import {
     renderRunestoneComponent,
     createTimedComponent,
-} from "../../common/js/renderComponent";
-import RunestoneBase from "../../common/js/runestonebase";
+} from "../../common/js/renderComponent.js";
+import RunestoneBase from "../../common/js/runestonebase.js";
```
bnmnetp, line 10: As far as I understand it there is no difference between including and not including the .js ??

But we should probably be consistent to reduce possible confusion.

bjones1: Based on https://webpack.js.org/api/module-methods/, I started using `.js` as the standard extension. (The code still works without it, but I think it's best to follow this recommendation.)

PR #1188, opened 2021-04-26
Add: Improve docs.

bjones1:


runestone/fitb/__init__.py
```diff
@@ -1 +1,4 @@
+# *********
+# |docname|
+# *********
```
bnmnetp, line 3: In my opinion, including stuff like this just creates noise in the documentation.  Additional headers that add nothing.

I wish we could come up with a metric for what is useful and what makes it harder to navigate and learn what you want to learn.

bjones1: That's a good point. How about ignoring `__init__.py` in Sphinx if it's empty? (I don't know of an automatic way to do this, but I can manually add these excludes for now.)

I would also like to come up with a documentation style guide to help us this about things like this. Here's my [current list](https://codechat.readthedocs.io/en/master/docs/style_guide.cpp.html) -- what would you add/change?

bjones1: I went ahead and added ignores for `__init__.py`.

bjones1: I implemented this with a short script in the BookServer project. If you like that. I'll do the same here.

bjones1: For now, I removed the additions above. Eventually, we can use the script to programmatically ignore the noise.

bjones1:
@bnmnetp , would you re-run these tests? Looks like a random failure.

runestone/fitb/fitb.py
```diff
@@ -117,21 +121,9 @@ def depart_fitb_node(self, node):
     self.body.remove(node.delimiter)
 
 
+# See the :ref:`docs <fill in the blank docs>` for the syntax of this component.
 class FillInTheBlank(RunestoneIdDirective):
     """
-    .. fillintheblank:: some_unique_id_here
```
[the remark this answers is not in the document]
bnmnetp: Indeed, the overview docs are better, but the docstrings allow us to programmatically access some simple documentation and include it in places like the web interface for writing questions.  I don't doubt there are other solutions, maybe better, but removing these docs breaks functionality that people are currently using.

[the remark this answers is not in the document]
bjones1: OK, restored a minimal example per Slack conversation.

PR #1194, opened 2021-05-06
SQLite fixes

bjones1:


bjones1:
It looks messy, but it's just a rebase onto latest master (I need newer commits to get all the testing updates I did) plus one commit to fix SQLite problems.

bjones1:
The resolve needed is trivial -- just adding one new setting.

PR #1198, opened 2021-05-12
Add: dynamically load JavaScript for each component.

bjones1:
This PR makes all Runestone Components dynamically loaded. The basic approach:

- When the document is ready, search for all HTML entities with a `data-component` attribute (which only occurs for the Runestone components). Use the value of this attribute (which is the name of the Runestone component) to dynamically load code for that component.
- In `bookfuncs.js`, send a "pre-login-complete" trigger/event instead of the "login-complete" event.
- In `webpack.index.js`, delay the `runestone:login-complete` trigger/event until dynamic loads are done and the "pre-login-complete" event has triggered.
- In `renderComponent.js`, when rendering a component, import its JS first.

I upgraded to webpack 5, so be sure to run a `npm install` then a `npm run build` to update everything when testing.


runestone/common/js/bookfuncs.js
```diff
@@ -258,10 +258,8 @@ function setupNavbarLoggedOut() {
 $(document).bind("runestone:logout", setupNavbarLoggedOut);
 
 function notifyRunestoneComponents() {
-    // Runestone components wait until login process is over to load components because of storage issues
-    $(document).trigger("runestone:login-complete");
-    if (typeof $pjQ !== "undefined")
-        $pjQ(document).trigger("runestone:login-complete"); // for parsons components which are using a different version of jQuery
+    // Runestone components wait until login process is over to load components because of storage issues. This triggers the `dynamic import machinery`, which then sends the login complete signal when this and all dynamic imports are finished.
+    $(document).trigger("runestone:pre-login-complete");
```
bnmnetp, line 262: I don't see anywhere in here where you trigger a runestone:login-complete, yet there are many places in the code waiting for that event that are not changed int his PR.

bnmnetp: NM, pulled the code locally and found it.

bjones1: Good deal!

runestone/common/js/renderComponent.js
```diff
@@ -18,45 +18,48 @@ export function renderRunestoneComponent(componentSrc, whereDiv, moreOpts) {
...
-    if (opt.orig) {
-        opt.lang = $(opt.orig).data("lang");
-        opt.useRunestoneServices = true;
-        opt.graderactive = false;
-        opt.python3 = true;
-        if (typeof moreOpts !== "undefined") {
-            for (let key in moreOpts) {
-                opt[key] = moreOpts[key];
+    // Import the JavaScript for this component before proceeding.
+    runestone_import(componentKind).then(() => {
```
bnmnetp, line 22: Lazy Question: could this be done as an await for consistency with the our use of async/await  in the rest of the components?

bjones1: It's not an `async` function, so I thought it would be easiest to use a `.then()` instead.

bjones1: I mean the `export function renderRunestoneComponent(componentSrc, whereDiv, moreOpts) {` which contains this code.

bnmnetp: But this change gives it async behavior yes?

I like to declare these functions async to avoid future mistakes thinking that I can call the function and expect synchronous results.

runestone/__main__.py
```diff
@@ -150,10 +150,10 @@ def build(all, wd):
     else:
         os.chdir(findProjectRoot())
     sys.path.insert(0, os.getcwd())
-    if not pathlib.Path(resource_filename("runestone", "dist/runestone.js")).exists():
-        click.echo("Error -- You are missing runestone.js please make sure")
-        click.echo("you have runestone installed correctly.")
-        click.echo("In a development environment run npm run build")
+    if not pathlib.Path(resource_filename("runestone", "dist/webpack.bundle.js")).exists():
+        click.echo("Error -- You are missing webpack.bundle.js.  Please make sure")
```
bnmnetp, line 154: Could we make this runestone.bundle.js - I am worried that we will run into name conflicts with a generic webpack.bundle.js name.  It will also make more sense for projects like PreTeXt to know that this is the runestone bundle.

bnmnetp: Also I assume that on the server side where we now include runestone.js we will change it to the bundle correct??

runestone/common/js/renderComponent.js
bjones1: True -- the component will sometimes be rendered later instead of synchronously. Would you like me to update that?

bnmnetp (commented):
Overall this is GREAT!  Thanks.

Just a couple of questions to resolve and I think we should merge as there are other new components in the works that should be done using this framework.

runestone/common/js/renderComponent.js
bnmnetp: That would be great.

I'm not sure we are 100% compliant with that rule, but we are getting closer, and I can add that to the CONTRIBUTING file after this is merged so future contributors will keep that in mind.

bjones1: OK, done!

runestone/__main__.py
bjones1: Yes, plus we need to include a few more files -- everything listed in `runestone/dist/webpack_static_imports.json`.

I'm looking at docs on how to do the rename.

bjones1:
Thanks! I think this will be a big step forward.

runestone/__main__.py
bnmnetp: Hmmm... `webpack_static_imports.json` is a file that gets generated by the `npm build` command.  What mechanism are we to use to get those loaded on the instructor interface pages??

bjones1: I think that's reasonably straightforward, since web2py has templates -- just read the JSON file for any of the books then transform this to `<script>` tags. Here's a crude sketch:

``` Python
{{ with open("path/to/webpack_static_imports.json", encoding="utf-8") as f:
    for js_import in json.loads(f.read()):
      response.write(f"<script src='_static/{js_import}'></script>\n")
}}
```

PR #1206, opened 2021-06-11
Add Parsons distractor helptext

vqum:
The `:adaptive:` option in Parsons problems renders a "Help Me" button that will assist the student by removing distractor blocks marked with `#distractor` or `#paired` in the Parsons problem's rst code. However, the student does not always understand why the distractor is incorrect. This change is meant to address this by allowing explanatory helptext to be added to distractor Parsons blocks in the problem's rst: `#distractor: explanatory helptext here` or `#paired: explanatory helptext here`. This was done by adding a Bootstrap tooltip (https://www.w3schools.com/bootstrap/bootstrap_tooltip.asp) to the distractor block after it is removed using the 'Help Me' button, where the tooltip appears upon hovering the mouse over the removed block.

This pull request is in conjunction with RunestoneServer pull request https://github.com/RunestoneInteractive/RunestoneServer/pull/1670 which adds the Boostrap tooltip functionality to doAssignment.html

For example:
```
.. parsonsprob:: testdistractortext
    :numbered: left
    :practice: T
    :adaptive:

    Log some text to the console.
    -----
    console.log("a");
    =====
    console.log("b");
    =====
    console.log("c");
    =====
    console.log("C"); #paired: Capital letter doesn't match the pattern.
    =====
    console.log("d");
    =====
    console.log("D"); #paired
    =====
    console.log("this one is a distractor"); #distractor: Told you so.
    =====
    console.log("Another distractor"); #distractor:
```

Above problem initially renders
image

This removed distractor block does have Boostrap tooltip added and therefore shows helptext upon mouse hover
image
image

Next removed distractor block does not have Bootstrap tooltip added because it was not given helptext
image

Next removed distractor block does have Boostrap tooltip added and therefore shows helptext upon mouse hover
image

Last removed distractor block does not have Bootstrap tooltip added because it was not given helptext
image

------------------------------------------------------------------------

Please let me know your thoughts.

I know this approach doesn't provide any visual indicator that the removed distractor block has additional interactivity - the original thought was to possibly add a button to removed distractor blocks that more clearly suggests interactivity, but I couldn't figure out how to get around the blocks being disabled upon removal.

Also, I think the W3Schools examples of Boostrap tooltips has additional CSS modifications that make theirs visually nicer. Their tooltip appears more or less immediately upon mouseover whereas mine seems to require pausing the mouse for a little bit of time before the tooltip appears. Additionally, their tooltips don't disappear when moving the mouse around the element, whereas mine does seem to disappear.


barbarer:
Love it!

PR #1211, opened 2021-06-24
Clean: removed unused config values.

bjones1:
More clean-up tasks: make the configuration for a book simpler where possible. In this case, remove unused or redundant settings:

- The `courses` table has the `login_required` flag, and it's been moved there. Remove it from the Sphinx book config.
- Likewise, `logLevel` is used once in the JS: `if (eBookConfig.useRunestoneServices && eBookConfig.logLevel > 0)`. Remove it, since `useRunestoneServices` can be examined instead of looking at both.

### DIFF
--- a/runestone/common/project_template/pavement.tmpl
+++ b/runestone/common/project_template/pavement.tmpl
@@ -34,8 +34,6 @@ options(
         outdir=serving_dir,
         confdir=".",
         template_args={
-            'login_required': '%(login_req)s',
-            'loglevel': %(log_level)d,
             'course_title': project_name,
             'python3': '%(python3)s',
             'dburl': '%(dburl)s',
--- a/conf.py
+++ b/conf.py
@@ -158,9 +158,9 @@ rst_prolog = (
 # requires **all** the following:
 #
 # - The use of Runestone services (``eBookConfig.useRunestoneServices === true``)
-# - Logging enabled (``eBookConfig.logLevel > 0``)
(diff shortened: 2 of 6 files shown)


bnmnetp:
The intention of `loglevel` was to provide some flexibility in what may or may not get logged beyond a simple binary yes or no.

bjones1:
Given that it's not used in the code currently, how about removing it until we have a need? When needed, adding a field to the DB would be a more flexible way to insert logging.

runestone/common/project_template/pavement.tmpl
```diff
@@ -34,8 +34,6 @@ options(
         outdir=serving_dir,
         confdir=".",
         template_args={
-            'login_required': '%(login_req)s',
```
bnmnetp, line 37: I'm not in favor of removing `login_required` This is still a very common use case for other individuals hosting runestone books on a simple server.  AND runestone.academy must still support the use of non-login books as base courses.  Every book on academy must have a free to read without login version.

bjones1: Effectively, `login_required` is a duplicate of `useRunestoneServices` -- we should eliminate one!

bnmnetp: I don't think they are duplicates.  Today, we still use Runestone services even if not logged in.  Actions are just logged anonymously.

bjones1:
Hmm, closing this one then.

PR #1217, opened 2021-07-10
Template cleanup

ascholerChemeketa:
A bunch of small edits focused on valid modern HTML and accessibility:
- Change attr quotes to "
- Remove duplicative/orphaned tags and unneeded attrs
- Set lang on HTML using pavement.tmpl
- aria-label for page progress

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -43,7 +43,7 @@
             crossorigin="anonymous"
         />
 
-        <script type="text/javascript">
+        <script>
             eBookConfig = {};
 
             eBookConfig.useRunestoneServices = false;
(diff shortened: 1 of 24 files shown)


bnmnetp:
Thanks I appreciate the PR.

The unit tests are failing for the codelens component. I can reproduce it consistently with

```
cd runestone/codelens/test
runestone build --all serve --port 8000
```

Now bring up `localhost:8000/` in your browser and you will get several assertion errors.  This does not happen on master, just on this PR.

ascholerChemeketa:
OK, working on that - something odd is going on. When I serve on windows and browse in Windows I do not get the failed assertions. When I serve from linux VM and browse in Windows, I do. Will get it figured out and then update.

PR #1219, opened 2021-07-16
Defer js loading to improve page time till first content

ascholerChemeketa:
Patch to substantially improve render time of pages.

Improves First Contentful Paint time as measured by Lighthouse in Chrome Dev tools by ~50% by setting all the runestone/sphinx js external <script> tags to load with `defer`. This guarantees that they are all loaded in listed order, but removes them as blocking elements for rendering the page. 

Implemented to be disabled by default. To enable behavior in a book, you must set "html_defer_js = True" in the conf.py of the project.

There are three sets of js files that are handled in slightly different ways:
- Files that are added from modules with add_js (or add_autoversioned_js) now explicitly set the defer attribute and add_autoversioned_js has been modified to pass it on. So any future model can opt out of having its scripts deferred if necessary.
- Bundles made by webpack have defer added to them programatically. 
- The jquery/underscore/docutils files that Sphinx adds have defer added to them programmatically.

The changes could be consolidated somewhat if we just blindly added `defer` to all scripts in script_files before render. But that would rule out particular scripts being added without defer by later code.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -319,6 +319,12 @@ html_show_sourcelink = False
 # This is the file name suffix for HTML files (e.g. ".xhtml").
 # html_file_suffix = None
 
+# It True, sets js files from Sphinx & Runestone to be loaded with defer attr
+# substantially speeding up page rendering. May cause issues with books that
+# have custom directives or raw html that assume jquery or another library
+# is loaded before body is parsed. 
+html_defer_js = True
+
 # Output file base name for HTML help builder.
 htmlhelp_basename = "PythonCoursewareProjectdoc"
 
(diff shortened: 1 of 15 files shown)


bjones1:
This looks great to me! If it works well, I'd suggest removing the config var and having it always enabled.

bnmnetp:
@bjones1 we decided on using the config var as there are a number of books out there that use

```
.. raw:: html

    <h1>my nifty hack</h1>
   <script>
    $(jquery).html....
   </script>
```

All of these kind of insertions into a runestone book will fail with deferred loading.  Having the switch allows us to control the migration this direction and quickly go back if we run into problems.

bnmnetp (approved):

PR #1220, opened 2021-07-20
Add: Files to support a wavedrom diagram.

bjones1:
This PR adds the ability to embed timing diagram in a Runestone book. It uses the dynamic loading system, so there's no JS burden unless it's used on a page. Here's an example (which this PR builds as part of the test code):

image

I'll add docs to the Overview book after this is merged.


bjones1:
OK, this is ready now.

PR #1221, opened 2021-07-20
Template cleanup

ascholerChemeketa:
Revised and reduced version of earlier PR.

- Fixes rendering bug caused by multiple class attrs on some elements.
- Removes redundant and unwanted meta tags from layout
- Removes extraneous div end tag from layout
- Removes type attr from scripts

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -43,7 +43,7 @@
             crossorigin="anonymous"
         />
 
-        <script type="text/javascript">
+        <script>
             eBookConfig = {};
 
             eBookConfig.useRunestoneServices = false;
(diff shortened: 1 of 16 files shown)


bnmnetp:
@ascholerChemeketa I *think* that the three conflicts are pretty simple to resolve, caused by merging your defer pr I'm guessing...

ascholerChemeketa:
Rebased to current master

bnmnetp (approved):

bnmnetp:
Unfortunately I had to undo the moves of the visually hidden.  They need to stay on the span containing the aria label or it ends up hiding the icon and the label and there is nothing in the navbar to click on.

PR #1222, opened 2021-07-20
Accessibility fixes

ascholerChemeketa:
- Adds label for on page progress bar
- Adds template arg to make sphinx set lang attr of HTML element on new projects
   - Existing books will need to have pavement.tmpl updated to get a lang


bnmnetp:
Just curious if you are doing these cleanups manually or if there is a tool we should add to our workflow somehow.

ascholerChemeketa:
Sorry, I don't have any perfect tools to add to the workflow. The changes I submitted were based on a combo of seeing things that were obviously broken on the page, the Lighthouse tab in Chrome Dev Tools, Validity chrome plugin, and the html linter (HTMLHint in VSCode).

bnmnetp:
Thanks for doing this.  Keeping up with accessibility is a never ending task and probably something we should develop some tests or workflow for so that we are keeping up and improving as we are adding features and fixing bugs.

runestone/common/project_template/pavement.tmpl
```diff
@@ -50,6 +50,7 @@ options(
             # These are used for non-dynamic books.
             'appname': 'runestone',
             'course_url': master_url,
+            'language': 'en',
```
bnmnetp, line 53: Where is this used?

ascholerChemeketa:
The language is getting passed through to the Sphinx package. It gets used on sphinx\themes\basic\layout.html:112
`<html{% if not html5_doctype %} xmlns="http://www.w3.org/1999/xhtml"{% endif %}{% if language is not none %} lang="{{ language }}"{% endif %}>`

runestone/common/project_template/pavement.tmpl
bnmnetp: gotcha -- just wanted to make sure it was not conflicting with any of our other uses of language

ascholerChemeketa:
I realized the 'language' setting I suggested for pavement.tmpl is not needed if language is set in conf.py.  But conf.tmpl defaults language to None and doesn't indicate that failing to give it a valid value will affect the HTML. So I've amended the PR to have the conf.tmpl set a valid value for language. 

If you don't want to have projects default to "en" for language, another alternative would be to just document the importance of setting that to a meaningful value.

bnmnetp:
english is a good default based on our readership….  There are definitely other sites like runestone.academy with books in other languages.  I guess this change will not effect current active books, until authors update their conf.py file.  

We could update the runestone init command to ask for the language and then set conf.py to that.

PR #1229, opened 2021-07-27
Change to SQL activecode - blocks using the same dburl (on same page) now share same db instance

cpainterwakefield:
Hopefully the code is pretty self-explanatory.  Previously, the code was saving the retrieved db file data and then creating a new database object on it each time.  Now, I save the created database object and re-use it each time.


bnmnetp:
@cpainterwakefield -- I'm just waiting to hear back from Libby as I wanted to make absolutely sure this won't break her book. -- 

If it does then we might need to add a flag to the directive to enable this new behavior.  Or maybe just a configuration parameter that we could add to the conf.py to disable the feature.

I really do think this is the right way to go moving forward I just don't want to create a bunch of work for someone a few weeks before classes start.

PR #1230, opened 2021-07-27
Allow select toggle preview pane to float in textbook pages

vqum:
https://github.com/RunestoneInteractive/rs/issues/60


bnmnetp:
Thanks @vqum

PR #1232, opened 2021-07-28
Layout fixes

ascholerChemeketa:
Fix rendering issues shown here:
https://docs.google.com/document/d/1C1jYh6DZryit2MzTblDOhdRd8Z1S_R2uD7It4YF067Q/edit?usp=sharing

Also cleans up the approach used in subchapter.html. The newer version relies less on js magic to create new components and add styles/events.


bnmnetp:
Thanks, for this, cleaning up the next/ previous buttons for narrow screens is very nice.

I don't know if you intended do make this change, but the width of the main text area is not like it used to be, and I don't think I like this change.  It is certainly one that reasonable people can disagree about.  But here is what I mean:

Here is what we have had for years:
Screen Shot 2021-07-30 at 1 14 05 PM

And here is what it looks like with this PR:
Screen Shot 2021-07-30 at 1 13 52 PM

At one point some people with more knowledge than I in UI design and maybe even some Ed Psych knowledge suggested the smaller column of text, it is easier to read, less cognitive load etc.  Even without that, the margins look too small to my eye for this current iteration.

I can see how this is an improvement on narrow devices, but our analytics show that 90% of our traffic is still from desktop browsers.

ascholerChemeketa:
No, that was not intended. And I don't see that locally. Messaging you in slack to try to get a screenshot of browser dev console so I can see what I am missing.

bnmnetp:
OK, so we know the margin problem is due to Sphinx 4.x!

I wanted to ask you what you thought about having the Previous/Next buttons at both the top and bottom for narrow screens.  That seems like a bit much to me.  I think with them always being at the bottom is just fine.

ascholerChemeketa:
Agreed that the layout is better without top buttons. Removed them from the layout and simplified subchapter.html back to not supporting being included multiple times.

PR #1247, opened 2021-08-19
group submit

bnmnetp:
- create shell for group submit
- Beta1 of group submit


bnmnetp:
Here's what it looks like so far...

Screen Shot 2021-08-19 at 3 28 44 PM

If you try to select more than the given limit (4 is the default)

Screen Shot 2021-08-19 at 3 29 38 PM

If the "team submitter" Doesn't add themselves, I add them automatically and recheck the length.

barbarer:
Yay

PR #1249, opened 2021-08-21
Fix: Old CSS overrode the width of a sidebar's float from 40% to auto

bjones1:
The old CSS made `.. sidebar::` elements take the entire line, due to the `width: auto` statement. Removing allows floats to work correctly (the default Sphinx CSS uses `width: 40%`).

Sadly, there's a bit of noise in the diff, since VS Code removed end-of-line whitespace.

# Before
image

# After
image

### DIFF
--- a/runestone/common/css/runestone-custom-sphinx-bootstrap.css
+++ b/runestone/common/css/runestone-custom-sphinx-bootstrap.css
@@ -208,7 +208,7 @@ a {
         margin: 0 3px;
         border-left: 1px solid #f2f2f2;
     }
-    
+
     .navbar-fixed-top .container {
         padding: 0;
     }
@@ -766,11 +766,11 @@ div.flash {
 
 .navLink a {
     display: inline-block;
-    background-color: white; 
-    border-style:solid; 
-    border-color:lightgrey; 
-    border-width:2px; 
-    width:100px; 
+    background-color: white;
+    border-style:solid;
+    border-color:lightgrey;
+    border-width:2px;
+    width:100px;
     height:50px
 }
 
@@ -789,7 +789,7 @@ div.flash {
 }
 
 @media (max-width: 600px) {
-    
+
     .navLink {
         display: inline-block;
         bottom: auto;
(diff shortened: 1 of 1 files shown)


runestone/common/css/runestone-custom-sphinx-bootstrap.css
```diff
@@ -869,11 +869,15 @@ div.flash {
     background-color: #ffffff;
 }
 
-.container .section>div.sidebar {
+.container {
```
bnmnetp, line 872: Are we sure this is what you want?  This is now a very broad rule.

bjones1: Oops, thanks. I somehow confused `.container .section>div.sidebar` with `.container, .section>div.sidebar`. I'll fix this, but insert a nice quote from a [CSS rant](https://medium.com/@joseph0crick/the-awful-css-language-9605534ca6e): 

> In conclusion, my philological studies have satisfied me that a gifted person ought to learn the basics of HTML in ten hours, JavaScript in thirty days, and CSS in thirty years. It seems manifest, then, that the latter ought to be trimmed down and repaired. If it is to remain as it is, it ought to be gently and reverently set aside among the dead languages, for only the dead have time to learn it.

bjones1: Doing some digging, [Sphinx just defines](https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/themes/basic/static/basic.css_t#L322) `div.sidebar`. But I'll keep it the way it was to be on the safe side.

bnmnetp: I think most everything is inside a `.container` I think I'm more aware of it because of some cross-over issues with overly broad rules that are causing some issues between Runestone and PreTeXt.

bjones1: Neat, that's good to know. CSS often gets so disconnected from the HTML is operates on, so that determining the actual interaction between the two is difficult.

bnmnetp: The bigger issue is actually that Sphinx 4.0 changes the master layout.html and uses better semantic html elements for the document.  This breaks a lot of our existing css for the books. :-( That will be a big project for someone to take on before we can move to sphinx 4.x

bjones1: Ouch -- good news in the long term, but definitely a lot of pain to get the CSS fixed...

PR #1250, opened 2021-08-25
Fix: Add /ns prefix to ajax paths.

bjones1:
Add the /ns prefix to JS ajax queries.

TESTING ONLY: this doesn't work if the bookserver is running standalone, instead of under nginx. We need to think about some nice way to support both cases.


bnmnetp:
As long as we are in agreement, I will just add it to my next commit, or if you are in a position to be making changes before I do just let me know on slack.

bjones1:
Sounds good. I'll let you add it.

PR #1253, opened 2021-08-31
Add: Compress webpage output.

bjones1:
This is probably the biggest win in terms of reducing bandwith needed for the Runestone Components. Many web servers (including nginx -- `gzip_static on;` in a `http` block) support serving statically compressed files.

Uncompressed, the JS, CSS, and WASM is 7.0 M (`du -hc *.js *.css *.wasm`).
Compressed, the JS, CSS, and WASM is 2.1 M (`du -hc *.js.gz *.css.gz *.wasm.gz`)

We can also look at better compressions schemes (zopfli, etc.), but web server support isn't as consistent. This is an easy fix and a big win.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "license": "ISC",
     "devDependencies": {
         "copy-webpack-plugin": "^8.0.0",
+        "compression-webpack-plugin": "^6.0.0",
         "css-loader": "^5.0.0",
         "css-minimizer-webpack-plugin": "^3.0.0",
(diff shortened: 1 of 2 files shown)


bnmnetp:
Cool!

In my nginx.conf file I have `gzip on` which, I think, makes gzip compress things on the fly.  This will save a bunch of cpu cycles.

bjones1:
Great!

bnmnetp (approved):

PR #1265, opened 2021-10-20
Node 17

bjones1:
Fixes to support the new release of node.

Manually run `npm update` since this requires very new webpack.


bnmnetp:
For me this required manually installing several additional packages that did not have polyfills from earlier versions of node.
* sax
* buffer
* string_decoder

PR #1277, opened 2021-11-23
Subchap num fix

bjones1:
The current code contains a case that returns a subchapter number of None, which violates a DB constraint (and is wrong). Here's a fix plus some docs.

### DIFF
--- a/runestone/chapterdb/dbchapterinfo.py
+++ b/runestone/chapterdb/dbchapterinfo.py
@@ -13,13 +13,15 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
+# **************************************************************
+# |docname| - Record chapter info in the database during a build
+# **************************************************************
 
 __author__ = "bmiller"
 
-import os.path
+from typing import Sequence
 from collections import OrderedDict
 import docutils
-import pdb
 from sphinx.util import logging
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


bjones1:
@bnmnetp, would you review/merge this? It's needed to avoid db exceptions after switching to the new db constraints in the BookServer.

runestone/chapterdb/dbchapterinfo.py
```diff
@@ -13,13 +13,15 @@
...
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
+# **************************************************************
+# |docname| - Record chapter info in the database during a build
+# **************************************************************
 
 __author__ = "bmiller"
 
-import os.path
+from typing import Sequence
```
bnmnetp, line 22: Sequence is deprecated in Python 3.9

bnmnetp (approved):

runestone/chapterdb/dbchapterinfo.py
bjones1: Thanks. From what I can tell, the new replacement of `from collections.abc import Sequence` [requires Python 3.9](https://www.python.org/dev/peps/pep-0585/). Do we want to support Python 3.8 or earlier versions? We might want to think about some sort of policy. Per https://endoflife.date/python, Python 3.6 is (essentiall) at its end of life; Python 3.7 is EOL in 1 1/2 years; Python 3.8 is EOL in ~3 years.

PR #1278, opened 2021-12-02
Dynamic problems, part 1

bjones1:
This is the first PR in a series which implements dynamic problems in Runestone. This series will focus on client-side support only, though it does have all the foundations needed for future server-side support.

This PR contains documentation on the proposed approach and preparation for the core code by improving the test framework. It does not change existing FITB or Runestone Components operation.


bjones1:
Closing this for now, since it's out of date. I'll reopen after Friday's PreTeXt discussion.

PR #1282, opened 2021-12-10
Fix: Correct sending an undefined answer.

bjones1:
If a students clicks "check me" on an mchoice without selecting an answer, the code returns "undefined" instead of an empty string.

### DIFF
--- a/runestone/mchoice/js/mchoice.js
+++ b/runestone/mchoice/js/mchoice.js
@@ -394,8 +394,9 @@ export default class MultipleChoice extends RunestoneBase {
             if (buttonObjs[i].checked) {
                 given = buttonObjs[i].value;
                 this.givenArray.push(given);
-                this.feedbackString += `<li value="${i + 1}">${this.feedbackList[i]
-                    }</li>`;
+                this.feedbackString += `<li value="${i + 1}">${
+                    this.feedbackList[i]
+                }</li>`;
                 this.givenlog += given + ",";
                 this.singlefeedback = this.feedbackList[i];
             }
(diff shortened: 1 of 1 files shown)


runestone/mchoice/js/mchoice.js
```diff
@@ -533,7 +534,8 @@ export default class MultipleChoice extends RunestoneBase {
     }
 
     async logMCMFsubmission(sid) {
-        var answer = this.givenArray[0];
+        // If there's no answer provided (the array is empty), use a blank for the answer.
+        var answer = this.givenArray[0] || "";
```
bnmnetp, line 538: this seems like a good fix.

runestone/mchoice/js/mchoice.js
```diff
@@ -394,8 +394,9 @@ export default class MultipleChoice extends RunestoneBase {
             if (buttonObjs[i].checked) {
                 given = buttonObjs[i].value;
                 this.givenArray.push(given);
-                this.feedbackString += `<li value="${i + 1}">${this.feedbackList[i]
-                    }</li>`;
+                this.feedbackString += `<li value="${i + 1}">${
+                    this.feedbackList[i]
+                }</li>`;
```
bnmnetp, line 399: This block looks like it should be hand formatted.  Not sure what formatter thinks it is a good idea to do `${` on one line and `}` after it makes it really hard to see that it is jQuery.

bnmnetp (approved):

runestone/mchoice/js/mchoice.js
bjones1: I'll revert this. I use [Prettier](https://prettier.io/) to format JS.

bjones1: But also, this isn't JQuery, it's a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

bjones1: Here's the code before Prettier:

``` JavaScript
                this.feedbackString += `<li value="${i + 1}">${this.feedbackList[i]
                    }</li>`;
```

This seems worse than what Prettier did to it!

bjones1: So on second thought, let's leave this change in.

runestone/mchoice/js/mchoice.js
bjones1: Thanks!

PR #1302, opened 2022-03-16
Updates to CSS etc to support docutils 0.17

bnmnetp:
This is an initial attempt to update styling for docutils 0.17.1

The major change by docutils was to move from `<div class="section">` to using the semantic tag `<section>` for structuring the document.

I have taken a different approach to the top level styling in this PR.  This selector `.container .section>*:not(.section):not(.ac_section)` has been the bane of my existence for years!  It basically constrains the width of everything in the section to be a sensible reading width.  Leaving the various components feeling more or less cramped when they could really use more space.  I have replaced that with the following selector

```
#main-content section > p,
#main-content section > pre,
#main-content section > table,
#main-content section > ul,
#main-content section > ol,
#main-content section > dl,
#main-content section > h1,
#main-content section > h2,
#main-content section > h3,
#main-content section > h4,
#main-content section > .admonition,
#main-content section > .toctree-wrapper {
```

Naming the tags explicitly may not be the best and we may need to add more tags to it over time but it leaves us free to style the components at whatever width we want!  Using good discretion so that the resulting page doesn't look like a hodgepodge.

This is a branch based on the rs2ptx branch, so it depends on it a little. but we can accept rs2ptx without needing to accept this one.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -44,7 +44,7 @@ cryptography==36.0.1
     # via
     #   pyopenssl
     #   urllib3
-docutils==0.16
+docutils==0.17.1
     # via
     #   codechat
     #   myst-parser
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ codechat==1.9.0
     # via -r requirements.in
 cogapp==3.3.0
     # via -r requirements.in
-docutils==0.16
+docutils==0.17.1
     # via
     #   codechat
     #   sphinx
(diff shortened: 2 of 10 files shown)


bnmnetp:
@ascholerChemeketa What do you think of this approach?  My CSS is not elegant and maybe you know of a better way to gives us the flexibility I would like.

bjones1:
Is this still relevant, or did it effectively get merged already?

bnmnetp:
No I merged the PR from @ascholerChemeketa

bnmnetp:
He had a much better solution so disregard what I had proposed.

PR #1303, opened 2022-03-18
Fix: logCurrentAnswer is async!

bjones1:


bnmnetp:
Interesting... Was this causing a problem for you?  In neither case are we looking at the response. So I'm curious.

bjones1:
I used the response in the dynamic problems flavor of fitb questions, which is what led me to this fix. So, these recent PRs are prep  for upcoming dynamic problems -- I'm looking for places where PRs help Runestone and don't break anything to make the later PRs easier to digest.

bjones1:
This PR depends on #1309, which needs to be merged before this can pass.

bjones1:
@bnmnetp, this is ready for review.

bnmnetp:
OK, this seems pretty simple.

PR #1307, opened 2022-03-19
Adding new directive: horizontal Parsons problems

amy21206:
A working version of horizontal parsons problems that only supports SQL for now.


bnmnetp:
Could you please add some selenium tests like those for regular parsons problems so that we have unit tests?

amy21206:
> Could you please add some selenium tests like those for regular parsons problems so that we have unit tests?

Hi, I just added the unit test in the latest commit. Thanks!

PR #1311, opened 2022-03-23
Style docutil 17 adjustments

ascholerChemeketa:
Suggested changes to handle basic width formatting for docutils 17.1

### DIFF
--- a/runestone/activecode/activecode.py
+++ b/runestone/activecode/activecode.py
@@ -61,7 +61,7 @@ def setup(app):
 
 
 TEMPLATE_START = """
-<div class="%(divclass)s">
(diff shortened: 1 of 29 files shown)


ascholerChemeketa:
Not a 100% visual match for old styling.

Tries to keep two container widths - standard narrow width for content that is for reading and a wider width for content that wants more space like Parsons and active code.

Also allows for better overriding of default widths. Most directives now will accept `:class: full-width` which will cause something that otherwise would be narrow (like a note) to take on the wider width.

Tested with RunestoneOverview and RunestoneComponents (quick flip through random selection of pages).

bnmnetp:
Closing manually as it is merged.

PR #1312, opened 2022-03-23
Improve CI

bjones1:
Fix: Install the Runestone Components locally so the npm run build works

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -28,26 +28,22 @@ jobs:
               if: always()
 
             - name: Setup npm and build runestone.js
-              id: create-runestone-bundle
               run: |
                   npm install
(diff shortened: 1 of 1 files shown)


bjones1:
@bnmnetp, this is ready for review.

bnmnetp:
Was this causing some intermittent failures?  I'm just curious what prompted you to investigate and make this particular change.

bjones1:
A minor thing -- the `npm build` that was [executed by the tests](https://github.com/RunestoneInteractive/RunestoneComponents/blob/master/runestone/conftest.py#L56) didn't work, so the old CI had to manually run it again. This turns out to be more important for dynamic fitb testing, which requires a specific flavor of npm build. However, it's a good overall fix for tests even without dynamic fitb.

PR #1313, opened 2022-03-23
Fix: Record source, line info for use in lp directive.

bjones1:


bjones1:
@bnmnetp, yet another to review...

bjones1:
True -- however, the lp directive raised an exception when showeval was used, since it expects all Runestone directives to have a `line` key.

bjones1:
I think you still need to merge this -- I don't have write access.

PR #1315, opened 2022-03-24
Docs: Add proposal for dynamic fitb problems.

bjones1:
This defines the currently-implemented approach to dynamic fitb problems, and serves as a discussion point to examine the approach taken.

I'll keep this PR as a draft until the team agrees on this.


rbeezer:
Thanks.  I've read this through pretty carefully for this afternoon's discussion.  See you then!

PR #1325, opened 2022-04-13
Divid checker

bjones1:
I wrote a test (high-stakes assessment) with a space in a divid, then discovered this the hard way (students taking the test). Here's code to check divids.

### DIFF
--- a/runestone/common/runestonedirective.py
+++ b/runestone/common/runestonedirective.py
@@ -25,9 +25,10 @@ from docutils.parsers.rst import directives
 from docutils.parsers.rst import Directive
 from docutils.utils import get_source_line
 from docutils.statemachine import ViewList
-
 from sphinx import application
 from sphinx.errors import ExtensionError
+from sphinx.util import logging
+
 
 UNNUMBERED_DIRECTIVES = [
     #    "activecode",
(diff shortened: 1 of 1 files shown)


runestone/common/runestonedirective.py
```diff
@@ -261,10 +264,33 @@ def __init__(self, *args, **kwargs):
...
 
-        self.options['optclass'] = self.options.get('class', "")
+        self.options["optclass"] = self.options.get("class", "")
 
         self.explain_text = []
 
+    # Check for a `valid HTML5 divid <https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute>`_.
+    def validate_divid(self, divid):
+        if (
+            # Per the spec, a divid must not contain `whitespace <https://infra.spec.whatwg.org/#ascii-whitespace>`_ (see also `Python string escape sequences <https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals>`_).
```
bnmnetp, line 274: Additionally, for XML compatibility divids should not begin with a numeric character.

bnmnetp: We have a handful of exercises in existing books that are just something like 3_1, 3_2 -- these will need to be migrated to a new name.

bjones1: I'm adding this to the code...which is also producing some pain for me -- I used random alphanumeric strings as divids, so I have some fixup to do on my book as well.

bnmnetp (approved):

PR #1326, opened 2022-04-18
Fix: Update tabbed routines to check for valid a divid.

bjones1:


bjones1:
This test failure looks like an intermittent problem and is not related to this PR -- Python 3.8 tests pass without problems, helping confirm this.

bjones1:
Pinging -- this is ready.

bnmnetp (approved):

PR #1330, opened 2022-05-12
Add support for Proof Blocks

SethPoulsen:
Add support for a directed acyclic graph (DAG) based grader to the Parson's problem plugin, so that it can support Proof Blocks problems (https://www.proofblocks.org/).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
                 "handsontable": "7.2.2",
                 "jexcel": "^3.9.1",
                 "jquery-ui": "1.10.4",
-                "jsnetworkx": "^0.3.4",
                 "select2": "^4.1.0-rc.0",
                 "sql.js": "1.5.0",
                 "vega-embed": "3.14.0",
@@ -494,20 +493,6 @@
             "engines": {
                 "node": ">=8"
             }
-        },
-        "node_modules/babel-runtime": {
-            "version": "5.8.38",
-            "resolved": "https://registry.npmjs.org/babel-runtime/-/babel-runtime-5.8.38.tgz",
-            "integrity": "sha1-HAsC62MxL18If/IEUIJ7QlydTBk=",
-            "dependencies": {
-                "core-js": "^1.0.0"
-            }
-        },
-        "node_modules/babel-runtime/node_modules/core-js": {
-            "version": "1.2.7",
-            "resolved": "https://registry.npmjs.org/core-js/-/core-js-1.2.7.tgz",
-            "integrity": "sha1-ZSKUwUZR2yj6k70tX/KYOk8IxjY=",
-            "deprecated": "core-js@<3.4 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Please, upgrade your dependencies to the actual version of core-js."
         },
         "node_modules/balanced-match": {
             "version": "1.0.2",
@@ -2466,22 +2451,6 @@
             "version": "1.10.4",
             "resolved": "https://registry.npmjs.org/jquery-ui/-/jquery-ui-1.10.4.tgz",
             "integrity": "sha1-oJb+X04PKraaBYXPEEVYd/V1Br0="
-        },
-        "node_modules/jsnetworkx": {
-            "version": "0.3.4",
-            "resolved": "https://registry.npmjs.org/jsnetworkx/-/jsnetworkx-0.3.4.tgz",
-            "integrity": "sha1-HAAl35QgjOtcxZ5lj5qb9f709vQ=",
-            "dependencies": {
-                "babel-runtime": "^5",
-                "lodash": "^3.3.1",
-                "through": "^2.3.6",
-                "tiny-sprintf": "^0.3.0"
-            }
-        },
-        "node_modules/jsnetworkx/node_modules/lodash": {
-            "version": "3.10.1",
-            "resolved": "https://registry.npmjs.org/lodash/-/lodash-3.10.1.tgz",
-            "integrity": "sha1-W/Rejkm6QYnhfUgnid/RW9FAt7Y="
         },
         "node_modules/json-parse-better-errors": {
             "version": "1.0.2",
@@ -4699,11 +4668,6 @@
                 "node": ">= 8"
             }
         },
-        "node_modules/through": {
-            "version": "2.3.8",
-            "resolved": "https://registry.npmjs.org/through/-/through-2.3.8.tgz",
-            "integrity": "sha1-DdTJ/6q8NXlgsbckEV1+Doai4fU="
-        },
         "node_modules/timsort": {
             "version": "0.3.0",
             "resolved": "https://registry.npmjs.org/timsort/-/timsort-0.3.0.tgz",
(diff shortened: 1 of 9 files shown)


SethPoulsen:
@barbarer @bnmnetp I thought you might like to glance at what I have going so far before we meet next week.

There are a few big TODOs in here still, but I think I've made some progress in the right direction.

SethPoulsen:
@bnmnetp From my end, I think this is all ready to go. Feel free to review & test. 

Thanks!

SethPoulsen:
Thanks @rbeezer and @bnmnetp!

One thing I forgot to mention--we have another feature in PrairieLearn called "block groups", which adds an extra condition to the grading where a subset of blocks must appear contiguously in the answer. This is helpful for doing things like Proof by Cases.

see an example here: https://github.com/PrairieLearn/PrairieLearn/blob/5dddcfb5b5ccc302790905c6ef165559a26e15ca/exampleCourse/questions/demo/proofBlocks/question.html#L35-L60

If people are interested in this, I can look in to adding the feature here as well. I'm not sure how the syntax would work with the Runestone markdown files, but with PreTeXt markup it could obviously be done similarly to the PrairieLearn example.

bnmnetp:
@SethPoulsen I have been telling people that RST will soon come to the point where we will add new features to the components, but those features will only be available when you are writing in PreTeXt.  As far as I am concerned this could be one of those features.

The PR would need to update the javascript and the html representation but would not need to provide RST markup.

SethPoulsen:
Hi Rob, these are great questions. 
  
  In PrairieLearn, leaving the `depends` attribute off is equivalent to `depends=""`, so the line does not depend on anything. I think this makes the most sense. 
  
  > Are the depends inside a pl-block-group redundant

  No---they are important. Each pl-block-group can have an arbitrarily complex DAG inside it (See, for example the test case all the way at the bottom of this currently open PR if you want to see a more complex example: https://github.com/PrairieLearn/PrairieLearn/pull/5871/files)

Hopefully that answer's your questions. If not, I'm happy to come by the drop-in, but I'm busy the next couple of Fridays, so it would have to be in a couple of weeks.

Thanks!

bnmnetp:
@rbeezer and @SethPoulsen 

I think block groups are going to be important for code too.

I was just having a discussion with @ascholerChemeketa on slack and he gave the folowing example

```
def a(): #tag 0; depend:;
=====
    return 10 #tag 1; depends:0;
=====
def b(): #tag 2; depends:;
=====
    return 20 #tag 3; depends:2;
=====
print(a + b) #tag 4; depends1,3;
```

There is nothing to prevent

```
def a():
def b():
  return 10
  return 20
print(a + b)
```
as a correct answer without the block groups.

bnmnetp:
Maybe a syntax like:

```
~begin-group~  tag g1 depend:;
def a(): #tag 0; depend:;
=====
    return 10 #tag 1; depends:0;
=====
~end-group~
~begin-group~ #tag g2 depends:;
def b(): #tag 2; depends:;
=====
    return 20 #tag 3; depends:2;
=====
~end-group~
print(a + b) #tag 4; depends g1,g2;
```

Ugh, that is not pretty.

bnmnetp:
Or this idea from Andrew is a lot less ugly.

```
def a(): #tag 0; group: g1; depend:;
=====
    return 10 #tag 1; group: g1; depends:0;
=====
def b(): #tag 2; group: g2; depends:;
=====
    return 20 #tag 3; group: g2; depends:2;
=====
print(a + b) #tag 4; depends g1,g2;
```

ascholerChemeketa:
Here is a slightly more complicated sample with nested groups. 

For purposes of programming problems, I think the rst syntax could be simplified to something like below. `under` would imply both a dependency and define a group. That amount of info looks compatible with the  `group_belonging` needed by the Prairie Learn algorithm.

```
def a(): #tag 0;
=====
    print("In a") #tag 2; under: 0;
=====
    if(True): #tag 3; under: 0;
=====
        x = 10 #tag 4; under: 3;
=====
        y = 20 #tag 5; under: 3;
=====
    return x + y #tag 6; under: 0; depends: 2,3;
=====
def b(): #tag 7;
=====
    return 20 #tag 8; under: 2;
=====
print(a() + b()) #tag 9; depends: 0,7;
```

SethPoulsen:
Ah, these are very interesting examples! Most people I work with only do single-function code blocks in parson's problems, so I hadn't thought of this use case.

Some comments on people's suggestions: 
I think that the `under` shorthand syntax suggested by @ascholerChemeketa is cool, I'm all for making things shorter when possible, but it seems to not be as flexible. For example, what if there was a block group where multiple different lines could be the first line in the block, I don't see a way to express that. 

The markup suggested by @rbeezer is essentially what we do in PrairieLearn, so it looks good to me. 

To be honest, I can't think of a syntax for block groups in the rst markup that both looks pretty and gives full support to the features. Maybe a slight improvement to @bnmnetp original suggestion could be to collapse of the group delimiters onto the same lines as the block delimiters, something like: 

```
~begin-group~  #tag g1 depend:;
def a(): #tag 0; depend:;
=====
    return 10 #tag 1; depends:0;
===== ~begin-group~ #tag g2 depends:;
def b(): #tag 2; depends:;
=====
    return 20 #tag 3; depends:2;
===== ~end-group~
print(a + b) #tag 4; depends g1,g2;
```
I am also curious though about how long the rst markup is going to be supported.

bnmnetp:
@Seth - so as I understand it you have code for this, (except for the runestone markup) we just need to agree on some markup for RST ?

I'm good with doing something simple, like what @ascholerChemeketa suggested as the long term solution is really PreTeXt.   

I'm not sure what makes the most sense.  We are in a kind of middle ground of moving to PreTeXt but there are books that are not yet converted, and cannot be converted because we are waiting on some other component to move to PreTeXt. 

0. I think PreTeXt markup is as Rob suggests... -  the issue is how it ends up looking in the html
1. use `under` to keep the current implementation fairly simple.  Assuming its easy for Rob to generate.
2. use a more complicated markup which makes parsing more complicated in the js -- seems bad considering we want to ditch 
3. Use parsons as a test case for going straight to json.  We would still want to generate the json for backward compatibility but it probably would not be that hard?
  * I'm suddenly imagining a way of specifying a parsons problem that looks like:

```
.. parsonsprob:: uniqueid
    :asptx:
    
    <exercise label="number-theory-proof" adaptive="yes">
   <title>Early morning exercise</title>
   <statement>
       <p>Put the blocks in order.</p>
   </statement>
   <blocks>
       <block order="2">
           <p>Get Up</p>
       </block>
       <block order="3">
           <choice><p>Drink a shot of rum</p></choice>
           <choice correct="yes"><p>Eat Breakfast</p></choice>
       </block>
       <block order="1">
           <p>Brush your teeth</p>
       </block>
   </blocks>
   <hint>Drinking before noon is not recommended.</hint>
</exercise>
    
```

Generating the json is now easy...

SethPoulsen:
Sorry no, I haven't yet implemented block groups in the JavaScript grader. That will take some amount of work. 

Are you suggesting we just essentially stick the PreTeXt markup into the HTML and have that traversed by the javascript? This seems like a find solution to me.

rbeezer:
> Use parsons as a test case for going straight to json. We would still want to generate the json for backward compatibility but it probably would not be that hard?

Not sure if matters to me or not: but I think Parson's was the most complicated thing I had to interpret;  math/code, numbered/not, indented/not, and maybe more.  I would suggest multiple-choice as a good first test case.

PR #1333, opened 2022-06-10
Update: Move to Poetry for dependency management and packaging.

bjones1:
An alternative approach to fixing RunestoneInteractive/rs#37.


bnmnetp:
doesn't there need to be a PR for server that makes a dev version for this?

bjones1:
There will be server changes, but I though I'd first get this working then create a PR for the server (which should be pretty easy). It's worth evaluating this as a possible solution IMHO.

bjones1:
Still to do: check that packaging works correctly, then a PR for the server.

bnmnetp:
Here are my concerns with our current approach.  Sorry if I'm repeating myself.

0. I would love to be able to run `poetry build` followed by `poetry publish` to release a new components!
1. Getting the dependencies right is very tricky and error prone.   Often requiring multiple rounds of `poetry update` in either BookServer or RunestoneServer and a `pip-compile ...` in RunestoneComponents.  rinse repeat. If I have trouble getting it right how can we expect others to do so.  I don't want to be in the position of cleaning up PRs that work for other people but not for a production release.
2. The ONLY way to update said dependencies is via a shell command inside a running docker container.  I think this really constrains who will contribute.  We have to be new contributor friendly.
3.  We have recently seen that getting docker up and running is not easy even for an accomplished PhD student with strong hacking cred.  (I can only point the finger of blame at myself on that one). But many people are afraid of docker, think it is too much work to set up etc.
4. I still cannot efficiently do development on BookServer because I need to run `docker compose stop` and `docker compose up -d` in order for changes to be recognized.  `stop-servers` and `start-servers` does not do it for me.

Enough complaining....  Can we solve these problems, while adopting poetry across the board?

1. Can we construct a docker-tools subcommand that will reliably upgrade all of the required poetry.lock files so that we can have all 3 projects in sync?
2. Can we construct a docker-tools subcommand that will quickly and efficiently restart servers when we are doing development?

When we went down this current path late last year it was the best way forward to get onto docker and into a system where we could scale horizontally!  That was a HUGE step forward.  But it feels to me like there is another step that will decouple our various servers and use Docker better that will not get us in this tangle of reconciling dependencies across three projects.

bjones1:
> Here are my concerns with our current approach. Sorry if I'm repeating myself.
> 
> 0. I would love to be able to run `poetry build` followed by `poetry publish` to release a new components!  So yay for this!!

Great!

> 1. Getting the dependencies right is very tricky and error prone.   Often requiring multiple rounds of `poetry update` in either BookServer or RunestoneServer and a `pip-compile ...` in RunestoneComponents.  rinse repeat. If I have trouble getting it right how can we expect others to do so.  I don't want to be in the position of cleaning up PRs that work for other people but not for a production release.  If someone works on the Components and updates a dependency, it will likely cause breakage for me after I accept the PR and try to run it in an integrated system rather than the isolated unit tests we now use for components only development.  Note -- I think isolated development is a good thing if someone wants to build us a new javascript widget and then can write selenium tests for it then they should not have to worry about spinning up the entire ecosystem.

This is the core problem. Because of the way I run my server (installing my code from git as path dependencies, rather than using released versions), I rarely encounter these problems, so the following is all a bunch of guessing on my part. Please advise/correct/provide a failing case so I can better understand what's going wrong. But, working on theory:

There are two cases AFAIK. In the simplest case, we're installing in development mode (very close to how my production server runs). In this case, poetry gets all the loose dependency specifications for all three projects then solves. It takes a while, but works well in most cases. (I've seen some corner cases where the Runestone Component dependencies Poetry uses are out of date -- manually deleting something (`dist/`? `build/`? `egg-info/`? I don't remember) fixed this. I think moving to a Poetry-based project will help fix this problem.)

The case I'm guessing you encounter is production: you run `poetry update`, then `poetry build` and `poetry publish` for both the Bookserver and the Runestone Components. This freezes the versions of **all** their dependencies, and may even produce conflicting dependencies, since these two projects don't know about dependencies of their siblings. Then, running `poetry update` on the Runestone Server means it must try to satisfy this very narrow set of frozen dependencies, sometimes resulting in a solver failure.

Assuming this is the root problem, I have some ideas on how to fix this. **Let me know what you think.**

> 2. The ONLY way to update said dependencies is via a shell command inside a running docker container.  I think this really constrains who will contribute.  We have to be new contributor friendly.

Yes, this is definitely a challenge. For BookServer development or Runestone Components, it's quite possible to simply pip install then develop those. With the Runestone server, I don't see any way around this, but I'm always open to ideas.

> 3. We have recently seen that getting docker up and running is not easy even for an accomplished PhD student with strong hacking cred.  (I can only point the finger of blame at myself on that one). But many people are afraid of docker, think it is too much work to set up etc.

Aack. I'd hoped that my nifty script would help with this -- unless it's Windows, the script should auto-install Docker. I'd like to find more people willing to try this then see where things break so we can improve the process.

> 4. I still cannot efficiently do development on BookServer because I need to run `docker compose stop` and `docker compose up -d` in order for changes to be recognized.  `stop-servers` and `start-servers` does not do it for me.  On my M1 Mac Max with 64GB of memory it takes 12+ minutes to restart! -- most that is in the chmod commands. but even with QUICK_START set to Yes it is 3+ minutes.

Wow, that's terrible. Do you know what's missing / broken with `start-servers` and `stop-servers`? That would be the quick route from what I see.

> Enough complaining.... Can we solve these problems, while adopting poetry across the board?
> 
> 1. Can we construct a docker-tools subcommand that will reliably upgrade all of the required poetry.lock files so that we can have all 3 projects in sync?

I have some ideas (see above) -- let me know if my guess about the underlying failure cause is correct, so we can talk about how to fix this.

> 2. Can we construct a docker-tools subcommand that will quickly and efficiently restart servers when we are doing development?

See above (item 4) -- I don't know what's not working there, but would be happy to change/fix whatever needs it.

> When we went down this current path late last year it was the best way forward to get onto docker and into a system where we could scale horizontally! That was a HUGE step forward. But it feels to me like there is another step that will decouple our various servers and use Docker better that will not get us in this tangle of reconciling dependencies across three projects.

If we can fix these problems, it will definitely help a lot!

bnmnetp:
> > 1. Getting the dependencies right is very tricky and error prone.   Often requiring multiple rounds of `poetry update` in either BookServer or RunestoneServer and a `pip-compile ...` in RunestoneComponents.  rinse repeat. If I have trouble getting it right how can we expect others to do so.  I don't want to be in the position of cleaning up PRs that work for other people but not for a production release.  If someone works on the Components and updates a dependency, it will likely cause breakage for me after I accept the PR and try to run it in an integrated system rather than the isolated unit tests we now use for components only development.  Note -- I think isolated development is a good thing if someone wants to build us a new javascript widget and then can write selenium tests for it then they should not have to worry about spinning up the entire ecosystem.
> 
> This is the core problem. Because of the way I run my server (installing my code from git as path dependencies, rather than using released versions), I rarely encounter these problems, so the following is all a bunch of guessing on my part. Please advise/correct/provide a failing case so I can better understand what's going wrong. But, working on theory:
> 
> There are two cases AFAIK. In the simplest case, we're installing in development mode (very close to how my production server runs). In this case, poetry gets all the loose dependency specifications for all three projects then solves. It takes a while, but works well in most cases. (I've seen some corner cases where the Runestone Component dependencies Poetry uses are out of date -- manually deleting something (`dist/`? `build/`? `egg-info/`? I don't remember) fixed this. I think moving to a Poetry-based project will help fix this problem.)
> 
> The case I'm guessing you encounter is production: you run `poetry update`, then `poetry build` and `poetry publish` for both the Bookserver and the Runestone Components. This freezes the versions of **all** their dependencies, and may even produce conflicting dependencies, since these two projects don't know about dependencies of their siblings. Then, running `poetry update` on the Runestone Server means it must try to satisfy this very narrow set of frozen dependencies, sometimes resulting in a solver failure.
> 
Not really in production, this happens on my dev server when I try to prepare a new release FOR production.  Say I have made changes to BookServer, Components, and RunestoneServer over the course of a week or two of development.

1. run `poetry update` ,  `poetry build` and `poetry publish` to make a new BookServer to install in production.
2. run `pip-compile` and then run my `makeRelease` script to make a new released version of Components to install in production.
3. Run `poetry update` in the docker container to prepare for me to commit a new commit for the RunestoneServer .  This is the failure point!  And would certainly be a failure in production if I hadn't done this step.

> Assuming this is the root problem, I have some ideas on how to fix this. **Let me know what you think.**
> 

**Great** I would love to hear your ideas!

> > 2. The ONLY way to update said dependencies is via a shell command inside a running docker container.  I think this really constrains who will contribute.  We have to be new contributor friendly.
> 
> Yes, this is definitely a challenge. For BookServer development or Runestone Components, it's quite possible to simply pip install then develop those. With the Runestone server, I don't see any way around this, but I'm always open to ideas.
> 
> > 3. We have recently seen that getting docker up and running is not easy even for an accomplished PhD student with strong hacking cred.  (I can only point the finger of blame at myself on that one). But many people are afraid of docker, think it is too much work to set up etc.
> 
> Aack. I'd hoped that my nifty script would help with this -- unless it's Windows, the script should auto-install Docker. I'd like to find more people willing to try this then see where things break so we can improve the process.
> 
I **think** this is more of a perceptual problem and we just need to do more to give people confidence that it is all goign to work.

> > 4. I still cannot efficiently do development on BookServer because I need to run `docker compose stop` and `docker compose up -d` in order for changes to be recognized.  `stop-servers` and `start-servers` does not do it for me.  On my M1 Mac Max with 64GB of memory it takes 12+ minutes to restart! -- most that is in the chmod commands. but even with QUICK_START set to Yes it is 3+ minutes.
> 
> Wow, that's terrible. Do you know what's missing / broken with `start-servers` and `stop-servers`? That would be the quick route from what I see.
This is only for changes to BookServer...  So there is something happening in the initial startup maybe a `poetry build` or `poetry install` of  BookServer that does not happen on a simple stop/start servers sequence ????

bjones1:
> I think this is more of a perceptual problem and we just need to do more to give people confidence that it is all goign to work.

Sounds good. I hope we'll start to build more confidence at AIM. Another way to build confidence is providing good docs -- definitely a place where there's a lot to do.

bjones1:
> > > 1. Getting the dependencies right is very tricky and error prone.   Often requiring multiple rounds of `poetry update` in either BookServer or RunestoneServer and a `pip-compile ...` in RunestoneComponents.  rinse repeat. If I have trouble getting it right how can we expect others to do so.  I don't want to be in the position of cleaning up PRs that work for other people but not for a production release.  If someone works on the Components and updates a dependency, it will likely cause breakage for me after I accept the PR and try to run it in an integrated system rather than the isolated unit tests we now use for components only development.  Note -- I think isolated development is a good thing if someone wants to build us a new javascript widget and then can write selenium tests for it then they should not have to worry about spinning up the entire ecosystem.
> > 
> > 
> > This is the core problem. Because of the way I run my server (installing my code from git as path dependencies, rather than using released versions), I rarely encounter these problems, so the following is all a bunch of guessing on my part. Please advise/correct/provide a failing case so I can better understand what's going wrong. But, working on theory:
> > There are two cases AFAIK. In the simplest case, we're installing in development mode (very close to how my production server runs). In this case, poetry gets all the loose dependency specifications for all three projects then solves. It takes a while, but works well in most cases. (I've seen some corner cases where the Runestone Component dependencies Poetry uses are out of date -- manually deleting something (`dist/`? `build/`? `egg-info/`? I don't remember) fixed this. I think moving to a Poetry-based project will help fix this problem.)
> > The case I'm guessing you encounter is production: you run `poetry update`, then `poetry build` and `poetry publish` for both the Bookserver and the Runestone Components. This freezes the versions of **all** their dependencies, and may even produce conflicting dependencies, since these two projects don't know about dependencies of their siblings. Then, running `poetry update` on the Runestone Server means it must try to satisfy this very narrow set of frozen dependencies, sometimes resulting in a solver failure.
> 
> Not really in production, this happens on my dev server when I try to prepare a new release FOR production. Say I have made changes to BookServer, Components, and RunestoneServer over the course of a week or two of development.
> 
> 1. run `poetry update` ,  `poetry build` and `poetry publish` to make a new BookServer to install in production.
> 2. run `pip-compile` and then run my `makeRelease` script to make a new released version of Components to install in production.
> 3. Run `poetry update` in the docker container to prepare for me to commit a new commit for the RunestoneServer .  This is the failure point!  And would certainly be a failure in production if I hadn't done this step.
> 
> > Assuming this is the root problem, I have some ideas on how to fix this. **Let me know what you think.**
> 
> **Great** I would love to hear your ideas!

I have several ideas -- let me start with simplest: our goal (for the moment) is to use a single venv for Bookserver, the Runestone Components, and the Runestone Server. (Eventually, we'll factor out some of these dependencies, so each can run in its own venv per [this dicussion](https://github.com/RunestoneInteractive/RunestoneComponents/pull/1332#issuecomment-1150634952).)

The standard release process is to release the Bookserver and the Runestone Components to PyPI, then use these releases. This allows us to clearly define when the code is in a release state, allows others to download a specific version, and a host of other good things. This is also what causes the problems, since we're trying to maintain a single, unified venv while separately assuming (by the release process, which pins all dependencies) that each package can run in its own venv. (I wish Poetry would have an option to publish the loosest possible dependencies for a release package, instead of using the tightest possible dependencies.)

So, the assumptions in the current release process aren't true in our release process; we need to re-think what a release means. We want to build a version of the Runestone Server which works with a specific released versions of both the Bookserver and the Runestone Server, but with dependency resolution taking place during this build, rather than before the build (by the current separate release process for the Bookserver and the Runestone Components.) To do this:

1. Check out the correct commit / tag from the Runestone Server, Runestone Components, and Bookserver repos. In Poetry terms, then means that for our production build, we're still using path dependencies, **not** PyPI dependencies.
2. Perform dependency resolution in the Runestone Server, saving this to the usual `poetry.lock` file.
3. Make a release from this. Here's where it's less obvious:

    a. Ideally, we'd take this `poetry.lock` file pick out only the dependencies needed for the Bookserver, creating its `poetry.lock` from that, then do the same for the Runestone Components, creating a different `poetry.lock` for that. Then, publish these releases on PyPI. I have ideas on how to do this, but they rely on assumptions and will take some time.

    b. More practically, call this good enough: the Runestone Server `poetry.lock` file has everything we need!

Does that make sense? What are your thoughts?

bnmnetp:
I think I am basically doing what you describe to make a release now.  I should not have combined 3.a and 3.b in steps 1 and 2 when I wrote it up earlier.

1. Run `poetry update` , `poetry build` in  BookServer
2. Run pip-compile in Runestone Components
3. Run poetry update in the docker container to prepare for me to commit a new commit for the RunestoneServer. If all goes well then:
    a. Run `poetry publish` in BookServer
    b. Run my `makeRelease` script in Components
    c. Try to remember to update the VERSION File for RunestoneServer
    d. tag RunestoneServer
4. Otherwise:
    a. Figure out how to fudge whatever I need to in steps 1 and 2  to make step 3 happy.

The problem is I end up in 4.a too often and that makes me unhappy.


I do not think the best practice solution is to run each in its own venv.  In my opinion the target we should aim for is one in which we run bookserver and celery in their own Docker containers just like we do for JOBE, Postgres and Redis.  Then we look carefully at what we are importing from runestone that makes it a dependency. For example nowhere in *production* BookServer do we actually import anything from runestone. (with the exception of lp) But there are places in RunestoneServer where we do, I think those could be factored out somehow.... 

RunestoneServer needs to RUN `runestone build`  to create previews of problems written in the web interface, but we will soon be replacing that with an xsl transform.  But even Running `runestone build` could be done with the runestone code residing in its own container for `docker run runestone build`

bjones1:
Based on our conversation at the conference, I'm keeping this a simple conversion from the usual `setup.py` to Poetry. It should all work similarly as before. Testing with the overview book shows that the components works locally. (The audio tour was missing files, but it's the same on https://runestone.academy/ns/books/published/overview/ActiveCode/audiotours.html).

bnmnetp:
OK, I'm inclined to test this locally and then merge as a step forward.  But...

I found this example https://testdriven.io/blog/fastapi-and-celery/ to be full of inspiring ideas for how we might move forward.

bjones1:
Even though tests fail, I've marked this as ready:

- The failures in activecode and timed problems come from a partial (I'm assuming) rename in 152764ad -- `edList` becomes `componentMap`, which causes failures.
- The hparsons test fail; I have fixes for them in RunestoneInteractive/RunestoneComponents#1339...but these test fail as `urllib3` again rears its ugly, misversioned head:
`pkg_resources.ContextualVersionConflict: (urllib3 1.26.7 (/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages), Requirement.parse('urllib3==1.26.9'), {'runestone'})`.

So, I propose we merge this, then I'll rebase the hparsons fixes on top of it to get these passing. I assuming @bnmnetp will push fixes for the activecode and timed failures.

bjones1:
The fix to the Runestone server that should accompany this is a one-liner (in `pyproject.toml`):

```
runestone-dev = { path = "../RunestoneComponents-dev", develop = true }
```

bnmnetp:
Ugh, the `componentMap` changes leaked over from my overhaul compatibility branch.  They got lost  amongst the black updates. 😞 .  I will pull this back to the branch and undo it here.

bjones1:
> I found this example https://testdriven.io/blog/fastapi-and-celery/ to be full of inspiring ideas for how we might move forward.

Thanks! I read through and made a few tweaks to the Celery config. I don't yet run Flower locally, but could if you're interested. It does offer good insight into what's going on in the task queue.

bnmnetp:
I wasn't that interested in Flower, but I was very intrigued by how they structured the project and set up the docker-compose.yml file.

bjones1:
The option to move Celery in one or multiple containers is definitely helpful -- if we every have some long-running/high-load jobs, that's always an option for us. I use it to limit the number of simultaneous simulation jobs, since the simulator for the microcontroller I use takes about 750 MB of RAM. At that size, it's easy to run out of RAM and crash the server if there's not a way to queue up jobs and run a subset (three in my case) in parallel.

bnmnetp:
I am going to use celery for two things:

1. integrated with web2py I am going to use it when instructors want to download the entire log for their course.  The query and processing required for that take too long for a single http request.
2. I am building a very small, and separate server with fastapi to allow authors to pull from github, build and deploy their books.  This will run on my "staging" server where I build things now.  But will give them some control.

bjones1:
Neat! Do you want me to help with changing / updating how Celery works currently? (Run in multiple containers, etc.)?

I used to run Celery in web2py for the same purpose, but moved that to the BookServer in the port. Mine was simpler (it ran inside a single http request) since the limitation wasn't on the processing time, but on RAM used.

bjones1:
I also realized that my code doesn't currently work in Celery if Celery runs in a separate container -- the Celery workers load data from a temp file produced by web2py. I'll need to fix that first.

bjones1:
Now that e575c17d is committed, this PR plus RunestoneInteractive/RunestoneComponents#1339 (which depends on this PR) should get us back to all tests passing.

bnmnetp:
This seems to work pretty well, and is a step forward in simplicity I think.  It does not entirely eliminate my build script.

1. I still have to run `npm run dist` to make sure I have rebuilt all of the webpack stuff.
2. I have to git tag everything manually
3. I have to create the xml version of the webpack static files json file.

Definitely not show stoppers, but I wonder if there is a way to add some hooks to poetry that allow all of that to happen?

And I still want to be able to develop in "edit mode" like we have been doing.  I know that can be done.

bnmnetp:
Looks like the built in plugin system will work very nicely.

bnmnetp:
We are in a broken state now, I tried adding your one line to RunestoneServer but it is not working.  What else did you do to get that to work?

bjones1:
Working on it now -- I'm having trouble with Celery not running...

bnmnetp:
I think the `/proc/1/fd/1` trick is not working for celery for some reason.

bjones1:
Yes, I agree. The `www-data` user doesn't have write permission to it. I'll change that for now.

PR #1340, opened 2022-06-30
Updated the log book event in the Toggle feature

Xinying329:
Log book event is now available when:
- Click Help me button
- Receive pop-up alert
- Close Parsons problem

### DIFF
--- a/runestone/activecode/js/activecode.js
+++ b/runestone/activecode/js/activecode.js
@@ -217,6 +217,7 @@ export class ActiveCode extends RunestoneBase {
                         "2px solid #b43232"
                     );
                     this.isAnswered = true;
+                    // the first time the student types in the write-code box
                     this.logBookEvent({
                         event: "activecode",
                         act: "edit",
@@ -265,9 +266,7 @@ export class ActiveCode extends RunestoneBase {
         // The run is finished; re-enable the button.
         this.runButton.disabled = false;
         this.runCount += 1;
-        if (this.is_toggle && this.runCount == 3) {
-            alert("Help is Available Using the Toggle Question Selector");
-        }
+        this.toggleAlert();
     }
 
     createControls() {
@@ -281,7 +280,7 @@ export class ActiveCode extends RunestoneBase {
         $(butt).addClass("btn btn-success run-button");
         ctrlDiv.appendChild(butt);
         this.runButton = butt;
-        console.log("adding click function for run");
+        // console.log("adding click function for run");
         this.runButton.onclick = this.runButtonHandler.bind(this);
         $(butt).attr("type", "button");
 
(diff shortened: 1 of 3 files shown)


runestone/activecode/js/activecode.js
```diff
@@ -649,7 +648,7 @@ export class ActiveCode extends RunestoneBase {
             $(this.timestampP).text(`${curVal} - ${pos + 1} of ${outOf}`);
             this.logBookEvent({
                 event: "activecode",
-                act: "slide:" + curVal,
```
bnmnetp, line 652: I believe this is correctly logging that the history slider is changing positions.  Why are you changing the act to save&Run ??

runestone/activecode/js/activecode.js
```diff
@@ -1272,6 +1271,19 @@ Yet another is that there is an internal error.  The internal error message is:
         }
     }
 
+    async toggleAlert(){
```
bnmnetp, line 1274: I see that you have declared this method as async, yet you are not using `await` for anything inside the function and you are not `await`-ing its use on line 269 in activecode.js

If there is reason to wait for the results from logBookEvent then use await there, if there is not reason to await anything then perhaps this function should not be declared as async ?

runestone/activecode/js/activecode.js
```diff
@@ -1272,6 +1271,19 @@ Yet another is that there is an internal error.  The internal error message is:
...
+        if (this.is_toggle && this.runCount == 3) {
+            if (this.errinfo != "success" || this.unit_results.substring(8,11)!=100.0){
+                setTimeout(function() {alert("Help is Available Using the Toggle Question Selector! You can try the Mixed-up Question first."); }, 500);
```
bnmnetp, line 1277: Why the delay of 500 msec before you show the alert??

runestone/parsons/js/parsons.js
```diff
@@ -2042,8 +2042,10 @@ export default class Parsons extends RunestoneBase {
         //this.helpButton.disabled = true;
         //}
         // if less than 3 attempts
+        console.log("msg_parson_atleast_three_attempts")
```
bnmnetp, line 2045: If these are for debugging then please remove them.

bnmnetp (changes requested):

runestone/activecode/js/activecode.js
Xinying329: I add this timer to show the code running results before the alert pops up, otherwise the alert will pop up immediately before seeing the actual test cases and error messages

runestone/parsons/js/parsons.js
Xinying329: Removed those lines

runestone/activecode/js/activecode.js
Xinying329: Turned it back to "slide"

PR #1341, opened 2022-06-30
Clean: Remove unnecessary file.

bjones1:
@ram8647, I don't think we need this file -- am I correct?


bjones1:
Note that test failures are due to urllib3 version conflicts, not to the contents of this PR. #1333 should fix the urllib3 problems.

ram8647:
That's right. I don't know how they got in there?

-- ralph

bjones1:
@ram8647, thanks for confirming!

PR #1343, opened 2022-07-04
Fix indentation handling for DAG based Parsons problems

ascholerChemeketa:
Indentation check code in line based grader assumes strict ordering. Break it into function for DAG grader to override.

DAG grader assumes every block has a tag and uses that to look up proper indentation in the given solution.

### DIFF
--- a/runestone/parsons/js/dagGrader.js
+++ b/runestone/parsons/js/dagGrader.js
@@ -95,12 +95,19 @@ export default class DAGGrader extends LineBasedGrader {
   checkCorrectIndentation(solutionLines, answerLines) {
       this.indentLeft = [];
       this.indentRight = [];
+
+      let indentationByTag = {};
+      for (let i = 0; i < solutionLines.length; i++) {
+        const line = solutionLines[i];
+        indentationByTag[line.tag] = line.indent;
+      }
+
       let loopLimit = Math.min(solutionLines.length, answerLines.length);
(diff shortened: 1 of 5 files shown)


ascholerChemeketa:
Checking on failed tests...

ascholerChemeketa:
Modified "tag" parsing code to allow multicharacter tags

Improved to handle non-sequentially numbered tags

Added a test and updated existing tests to remove deprecated constructs

ascholerChemeketa:
@bnmnetp I think this is ready... it is passing parsons tests.

hparsons tests are failing. It appears they are running on the Parsons test webpage instead of their own. I tried rolling back to a commit before my own changes and I am still seeing the same behavior. It looks like you are currently making changes in hparson... I don't *think* that the failed test is based on my code.

bnmnetp:
No, the failed test is a simple wording change.  For math problems that use parsons it does not make any sense to call the solution "program" so we agreed on the more general term of "answer"

bnmnetp:
Why do you think they are running on the Parsons test webpage?  When I run this locally I can see the page in the chrome browser and they are defilnitely horizontal problems.

SethPoulsen:
@ascholerChemeketa Thanks for catching this and fixing it. It was an easy thing for me to miss as I've only been working with proofs, not code most of the time.

PR #1344, opened 2022-07-06
Updated log book event in the toggle feature

Xinying329:
- Add await to logbookevent in the async method
- Remove debug lines
- Turn "save&run" back to "slide"


bnmnetp:
I have two general comments:

1. The reason to declare a function async is to let others know that the function is doing something asynchronously, AND that the function will return a Promise immediately.  Normally when you declare something async you will also be awaiting the result of some other async function inside your newly declared function.
2. I would like for there to be some consistency to our code.  using  jQuery`$(closeButton).click` and the equivalent pure Javascript form `closeButton.addEventListener("click", function...` in the same function is very confusing for future maintainers.  I would prefer to be moving toward pure Javascript.

Xinying329:
> 

Thanks! I have removed async and merged two click methods to one in pure javascript

PR #1348, opened 2022-07-10
move shared functions from rsmanage

bnmnetp:
@bjones1 

After not coming up with anywhere else to put these besides starting a whole different package I **think** this makes the most sense.  These functions are used by the rsmanage command  as well as the web interface for building and managing books.

If you have a different suggestion I'm happy to discuss....  I considered the following:

* code shared by bookserver, runestone server, and other new servers should go in runestone
* code shared by bookserver, runestone server, and other new servers should go a new runestone_common that we package separately from runestone components
* code shared by bookserver, runestone server, and other new servers should go in an rsmanage package

### DIFF
--- a/runestone/server/utils.py
+++ b/runestone/server/utils.py
@@ -155,7 +155,7 @@ def process_manifest(cname, mpath, click=click):
         )
 
 
-def check_project_ptx():
+def check_project_ptx(click=click):
     """
     Verify that the PreTeXt project is set up for a Runestone build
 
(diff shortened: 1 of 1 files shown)


bjones1:
This makes sense to me. However, I'd suggest adding comments / docs explaining that this code is shared across servers / repos.

Is the longer-term goal to move rsmanage to the BookServer? I assume it's a prereq to instructions being able to just pip install bookserver and run a SQLite-based server.

PR #1350, opened 2022-07-19
Make activecode components clean up old errors before displaying new ones

ascholerChemeketa:
Currently if a student is working in an activecode area, each attempt to build and run that has errors generates a new error box (each with identical html id) and displays it AFTER any existing error boxes created by previous attempts to build and run.

This deletes existing error boxes before displaying new ones.

An alternate approach would be to keep old errors, but display new errors on top where they are most visible and make it clear that the older boxes may no longer be relevant. I don't see much value in that unless the editor itself had a better sense of history.

### DIFF
--- a/runestone/activecode/js/activecode_js.js
+++ b/runestone/activecode/js/activecode_js.js
@@ -52,6 +52,9 @@ export default class JSActiveCode extends ActiveCode {
     }
 
     addErrorMessage(err) {
+        // Clear old errors for this box
+        let eContainerId = this.divid + "_errinfo";
+        $("#" + eContainerId).remove();
         // Add the error message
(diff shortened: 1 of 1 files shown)


bnmnetp:
Strange, I do not see the problem you describe.

ascholerChemeketa:
Sorry, I agree - looks like the code that is in runProg is doing this already. It was broken for me on a page, but I can't recreate it now. I think there must have been a character in the component's id that the js string template on line 1298 of activecode.js did not like.

bnmnetp:
I noticed you were changing `activecode_js.js` Maybe this is isolated to Javascript activecodes only?  I didn't check and am on vacation with pretty bad internet.

PR #1354, opened 2022-08-01
Fix: Only enable server-side grading if it's requested.

bjones1:
Fix for https://github.com/RunestoneInteractive/BookServer/issues/105.


bjones1:
I haven't tested this manually to verify server-side grading is turned off...just a quick first pass at a fix.

bnmnetp (approved):
This looks good.   I tested it with overview and it cleared out the feedback in the database and it is not trying to do server side grading.

bjones1:
Fantastic! Thanks for the testing. I'm glad this fixed the server-side grading bug.

Issue #1294, opened 2022-02-03
Adding Code Completion for active codes


bhoffman0:
It would be great to try out the code completion in the active code text areas (there should be an open source add-on in Code Mirror). This would really help students with learning disabilities and English as a second language learners in typing in their code. Thanks for anyone who wants to look into this!

bhoffman0:
It looks like the Code Mirror addon for code completion is fairly basic and only works for JavaScript:  https://codemirror.net/demo/complete.html. It might be possible to extend to other languages if it's just a list of keywords that it brings up. That might still be useful for students. The matchbrackets.js add-on would also be useful for students and easy to add:  https://codemirror.net/doc/manual.html#addons

bnmnetp:
Beryl,

There are a few other languages supported, html, css, javascript.  I found a python version as well.  

I think it would be a fun project for an advanced HS student or college student that wanted to see how Python or html is done and to write one for Java.  Maybe one of your teachers has a student looking for a real world project. Or it could be an option for one of Jan's students this summer.

I have enabled one called anyword which simply looks at the words in the current editor and does autocompletion on those.  Since lots of variables are reused many times in the code this is useful across all languages.

I have also enabled the matchbrackets.js that you mentioned.  Will release this with the saturday updates.

bnmnetp:
For reference if/when someone picks this up.   My simple changes to `activecode.js` can be enhanced by changing the autocomplete function from always loading anyword to something like this:

```
CodeMirror.commands.autocomplete = function(cm) {
    var doc = cm.getDoc();
    var POS = doc.getCursor();
    var mode = CodeMirror.innerMode(cm.getMode(), cm.getTokenAt(POS).state).mode.name;

    if (mode == 'xml') { //html depends on xml
        CodeMirror.showHint(cm, CodeMirror.hint.html);
    } else if (mode == 'javascript') {
        CodeMirror.showHint(cm, CodeMirror.hint.javascript);
    } else if (mode == 'css') {
        CodeMirror.showHint(cm, CodeMirror.hint.css);
    }
};
```
We will know the mode we want based on the parameters to the activecode directive.

bhoffman0:
Thank you for pursuing this! I'm excited to try out the new features this weekend. I'll arrange someone to work on it this summer.

dsanders12:
@bnmnetp @bhoffman0 I am interested in working on this project if no one has already been assigned.

bnmnetp:
[...]

PR #1355, opened 2022-08-02
Issue #1294

campbelle2:
This is the code we had when the code auto-completion began, but we are not sure why. If you figure it out or need us to change anything just let us know

### DIFF
--- a/runestone/activecode/js/activecode.js
+++ b/runestone/activecode/js/activecode.js
@@ -53,7 +53,7 @@ CodeMirror.commands.autocomplete = function(cm) {
     var doc = cm.getDoc();
     var POS = doc.getCursor();
     var mode = CodeMirror.innerMode(cm.getMode(), cm.getTokenAt(POS).state).mode.name;
-    console.log(mode)    
+    console.log(mode)   
     if (mode == 'xml') { //html depends on xml
         CodeMirror.showHint(cm, CodeMirror.hint.html);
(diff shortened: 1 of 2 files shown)


bnmnetp:
I looks like there are a few pieces missing?  I don't see where you have defined the keywords?

campbelle2:
I see what you mean. The file with the keywords is in the node_modules directory but I cannot view that directory on GitHib. Could it be something with that folder not appearing properly?

bnmnetp:
I think you need to add it locally and push to github.

campbelle2:
None of the changes I make in the node_modules folder appear in the source control tab of VS code so I cannot commit them. Do I need to do something specific to commit changes from that folder?

bnmnetp:
The node_modules folder is ignored.  Those are all installed automatically based on the contents of the package.json file when you run npm install.  In other words, those are third party libraries that you should not be modifying.

If you added files then you can move them to `activecode/js` and import them there.  If you modified other files I would be curious to know which ones.... There should be a way to do this without modifying third party libraries directly.

campbelle2:
All of the other modes/keyword banks for the CodeMirror autocompletion were in that folder so I assumed this one needed to be as well. The current import for the java mode in activecode.js is `codemirror/mode/java` so that may need to be changed if the java.js file needs to be outside node_modules.

PR #1361, opened 2022-09-04
Dropdown menu fixes

ascholerChemeketa:
Fix two minor but visually annoying layout issues in top of page dropdowns.

Padding of items inside social menu ("Runestone in social media", "Help support us", and Donate button)
Display of search button next to text input inside search menu

### DIFF
--- a/runestone/common/css/runestone-custom-sphinx-bootstrap.css
+++ b/runestone/common/css/runestone-custom-sphinx-bootstrap.css
@@ -341,7 +341,8 @@ div.container {
     margin-bottom: -6px;
 }
 
-.dropdown-menu > li > span {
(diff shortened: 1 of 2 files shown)


ascholerChemeketa:
Note for Brad - adding one more fix to the dark/light slider... still in progress.

ascholerChemeketa:
Added the Dark Mode fix. Think this is done now.

bnmnetp:
Thanks, I will get it reviewed in the next day or two.

bnmnetp:
Thanks!  the dark mode toggle looks much nicer!  I knew there had to be a way to do that.

I had not noticed that the search was broken either.  I copied that change to the server code so its fixed there now too... Thanks

PR #1362, opened 2022-09-08
Auto load history for activecode problems on page load

ascholerChemeketa:
Remove "Load History" button completely
Load historyscrubber on activecode problem initialization

### DIFF
--- a/runestone/activecode/js/activecode.js
+++ b/runestone/activecode/js/activecode.js
@@ -288,7 +288,7 @@ export class ActiveCode extends RunestoneBase {
             this.addDownloadButton(ctrlDiv);
         }
         if (!this.hidecode && !this.hidehistory) {
-            this.addHistoryButton(ctrlDiv);
+            this.addHistoryScrubber(true);
         }
         if ($(this.origElem).data("gradebutton") && !this.graderactive) {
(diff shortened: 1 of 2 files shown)


ascholerChemeketa:
Proposed changes for experiment with automatically loading history for activecode problems.

bnmnetp:
This looks good except for in a timed exam.  Yeah welcome to my world -- timed exams have always been problematic.  In a timed exam we show the slider from the start but only show them the first thing in the history.  Now we are getting double sliders.

Screen Shot 2022-09-08 at 4 56 46 PM

bnmnetp:
The overview book has a couple of timed exams you can play with.

bnmnetp:
I see ---

See the `timedInit` function in `TimedActiveCodeMixin` where I already have the call `await this.addHistoryScrubber(true);`

bnmnetp:
No need for you to do anything... I just removed that call and everything looks good.

I don't think there is any reason to worry about `awaiting` it when we load the page and are initializing everything.  I think I was just doing that for timed exams to be careful.

bnmnetp:
OK, this will be live on Saturday.  We will know a lot by Monday morning.

PR #1378, opened 2022-12-27
Removing unused dependencies in hparsons directive

zihan21206:
Updated the custom element used for horizontal parsons input and removed the unused dependencies, such as Quill.

Also used terser to reduce the code size of the custom element.


bnmnetp:
Hi,

The file that you call micro-parsons.js looks like a minified version of Sortable 1.14.0. ??

If that is the case then you should `npm install sortable` so that it is added and tracked through the package.json file.

Since we can't edit it, there is no reason for us to have it in our repo other than as a third party dependency.

zihan21206:
Hi Brad,

It is actually the compiled version of this repo:
https://github.com/amy21206/micro-parsons-element, which has sortablejs as
a dependency, but since sortable is the only one with a lisence that's why
it's the only readable text showing on top...

I will build it with sortable as a separate dependency we can import from
Runestone instead, but I was just trying to minimize the repo without huge
changes since we want to get it ready before the semester starts. But if
you think it is urgent, I will change it asap.

Thanks,
Zihan

bnmnetp:
OK, I see....

I think that what we want to work toward -- not right now, as we prepare for the semester -- is that the micro-parsons-element should be released/published using npm with its own versions.  Then in Runestone we install and manage the micro-parsons-element as a third party library.

It really does us no good to have a minified js file in our repo as we can't edit it.

We will need to document this so that when someone else needs to add something to micro-parsons they will know that part of the code is contained in a whole different repository.

amy21206:
Thanks for your help!

I added documentation in the js file for hparsons directory pointing edits to the micro-parsons-element repository, and I also created an issue (#1379) referring to this PR and addressing the future plan for replacing the minimized js file with the published package.

PR #1386, opened 2023-02-16
Add restoring local student answer for hparsons directive

amy21206:
Implemented checkLocalStorage(), setLocalStorage(), and restoreAnswers() for hparsons directive. 

Server implementation might be needed for restoring server-side answers. Documentation included in restoreAnswers().

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
                 "handsontable": "7.2.2",
                 "jexcel": "^3.9.1",
                 "jquery-ui": "1.10.4",
-                "micro-parsons": "https://github.com/amy21206/micro-parsons-element/releases/download/v0.1.3/micro-parsons-0.1.3.tgz",
+                "micro-parsons": "https://github.com/amy21206/micro-parsons-element/releases/download/v0.1.4/micro-parsons-0.1.4.tgz",
                 "select2": "^4.1.0-rc.0",
                 "sql.js": "1.5.0",
                 "vega-embed": "3.14.0",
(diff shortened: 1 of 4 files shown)


bnmnetp:
Working with localStorage but not with data from the server.

Here is the object that the server returns, but 

```
serverData
{id: 5, timestamp: '2023-02-23T01:55:22.625929', div_id: 'test_hparsons_block_1', sid: 'testuser1', correct: true, …}
answer: blocks: (4) ['SELECT', '*', 'FROM', 'test']
[[Prototype]]: Object
correct: true
course_name: "overview"
div_id: "test_hparsons_block_1"
id: 5
percent: 1
sid: "testuser1"
timestamp: "2023-02-23T01:55:22.625929"
```

Also, there is no reason to log the fact that we restored the answer.  None of the other components log that action.

bnmnetp:
@amy21206 I changed my code so that it now calls

```
this.hparsonsInput.restoreAnswer(serverData.answer.blocks);
```
But I'm still not getting a restore to work with server data.

Its hard to debug because I only have the minified version of the microparsons.  How do you have your development environment set up so that you can make changes to microparsons and then incorporate them into runestone without doing a release and npm update on the runestone side?

bnmnetp:
OK, I figured out a workflow....

I changed the imports in hparsons and copied everything from bin/ to the hparsons/js folder.

Now I am able to step through everything with the debugger.  And it appears that it is working....

I also re-updated the micro-parsons package with npm and with my small change it now appears to work either way....  So I don't think you need to do anything at the moment, but I would suggest eliminating the log message about restoring.

amy21206:
Sorry I just saw the message! I just changed my settings for github email notification so I can get the messages in time in the future. I will remove the log message about restoring, and let me know if there's anything else!

PR #1392, opened 2023-03-22
Python check

ascholerChemeketa:
Browser side python syntax help.

Branding question - what to call it in header:
"Syntax Tips"?
"Syntax Coach"?
"Syntax Warnings"?


bnmnetp:
I always thought the name "Code Coach" had a nice ring to it.  I'd be happy to resurrect that name.

bnmnetp:
They would all be orphans.  Definitely leftovers from an earlier era.  Server side support got cleaned out, but not  client side.

ascholerChemeketa:
Merged the other PR that cleaned up some activecode html/css into this as they were starting to overlap.

Cleaned out codecoach js and repurposed the div for python_check

image

PR #1393, opened 2023-03-22
Purge activecode layout cruft

ascholerChemeketa:
Gets rid of col-md-12 inside activecode components
Gets rid of clear divs that must have been used for side/side layout in activecode
Purges unused alignVertical code
Fixes margin under codelens


ascholerChemeketa:
Going to merge this into the Python check PR as the code is starting to overlap

PR #1394, opened 2023-03-24
New feature and bug fixes for hparsons directive

amy21206:
1. Feature: For SQL with execution-based feedback: allow adding hidden prefix and suffix to support visualization and grading of "UPDATE" and other statements that modify the table but do not return results;
2. Log update: Remove logs for restoring students' answers to be consistent with other directives;
3. Bug fix: Fix execution-based feedback for SQL always give correct as "T" when code fails to execute.

### DIFF
--- a/runestone/hparsons/hparsons.py
+++ b/runestone/hparsons/hparsons.py
@@ -102,10 +102,16 @@ class HParsonsDirective(RunestoneIdDirective):
         Here is the problem description. It must ends with the tildes.
         Make sure you use the correct delimitier for each section below.
         ~~~~
+        --hiddenprefix--
+        // code that is for scaffolding the execution (e.g. initializing database)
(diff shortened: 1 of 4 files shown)


amy21206:
Hi Brad @bnmnetp ,

Good morning! Could you please merging this PR and build the server before Tuesday? I am hoping to run a classroom study with the hparsons directive with execution based feedback on Tuesday and would require the ability to run some hidden code that is included in this PR.

Thank you,
Zihan

bnmnetp:
Hi @amy21206 and tagging @barbarer 

I have merged this, and rebuilt py4e-int with the new version. 

This will make your changes available through the web interface.  I have also updated the author server with this new version.

So you should be set for your experiment tomorrow.

Brad

PR #1396, opened 2023-04-07
Code coach modularize

ascholerChemeketa:
- Moved Pyflakes code coach to its own file.
- Update coach run/render code to more cleanly support multiple coaches. 
- Set up Code Coach string for internationalization.
- Fix bug where coach div not removed if all issues fixed

Coaches are added to a list during **ActiveCode** construction.

**runCoaches** function is triggered on program run. It asynchronously runs all coaches, but then waits for all results so Code Coach div can display any coach feedback deterministically in the order coaches were created. 

Questions/Comments:
- I could definitely see a case for more flexible addition of coaches to ActiveCodes instead of hard coding based on language in the constructor. I just don't have a clear idea of what that should look like.
- I *think* the internationalization is right...

### DIFF
--- a/runestone/activecode/js/activecode.js
+++ b/runestone/activecode/js/activecode.js
@@ -63,7 +63,6 @@ export class ActiveCode extends RunestoneBase {
         this.containerDiv = opts.orig;
         this.useRunestoneServices = opts.useRunestoneServices;
         this.python3 = true;
-        this.alignVertical = opts.vertical;
         this.origElem = orig;
         this.origText = this.origElem.textContent;
         this.divid = opts.orig.id;
--- a/runestone/activecode/js/activecode.js
+++ b/runestone/activecode/js/activecode.js
@@ -159,7 +159,7 @@ export class ActiveCode extends RunestoneBase {
         var linkdiv = document.createElement("div");
         linkdiv.id = this.divid.replace(/_/g, "-").toLowerCase(); // :ref: changes _ to - so add this as a target
         var codeDiv = document.createElement("div");
-        $(codeDiv).addClass("ac_code_div col-md-12");
+        $(codeDiv).addClass("ac_code_div");
         this.codeDiv = codeDiv;
         this.outerDiv.lang = this.language;
         $(this.origElem).replaceWith(this.outerDiv);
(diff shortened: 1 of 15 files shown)


ascholerChemeketa:
I stress-tested the deterministic output with a **TestCoach** that just waited a random amount of time before returning a set string. Works as intended when there are multiple coaches taking variable amounts of time.

ascholerChemeketa:
One aspect that might be annoying in existing books - Pyflakes complains about * imports.

Personally, I am using this as an excuse to replace them all in my book. But it would not be too hard to filter them out in the Pyflakes coach.

image

runestone/activecode/js/activecode.js
```diff
@@ -1222,6 +1227,35 @@ Yet another is that there is an internal error.  The internal error message is:
...
+
+        //clear anything after header in codecoach div and hide it
+        $(this.codecoach).children().slice(1).remove();
+        $(this.codecoach).css("display", "none");
+
+        //get code, run coaches
+        let code = await this.buildProg(false);
+        let results = [];
+        for(let coach of this.codeCoachList) {
+            results.push(coach.check(code));
```
bnmnetp, line 1241: Should there be a method to add a coach?  If I'm thinking about the future and adding a new coach I would want to have a method to do that so I don't have to modify this code...

ascholerChemeketa: No one should have to touch that code. 

Coaches are added up in the constructor (line 143). I'm not sure where else they would be injected from - that was one of my questions. I don't have a good sense of how to make creating coaches more modular.

runestone/activecode/js/activecode.js
```diff
@@ -137,6 +139,12 @@ export class ActiveCode extends RunestoneBase {
             this.caption = "ActiveCode";
         }
         this.addCaption("runestone");
+
+        //Setup CodeCoaches - add based on language
```
ascholerChemeketa, line 143: Coach creation

runestone/activecode/js/activecode.js
ascholerChemeketa: Hrm, can't figure out how to quote in those lines without starting a new conversation.

bnmnetp: I was imagining a method on the Activecode instance, but then I'm not sure what other object would call it, so having it in the constructor is fine for now.

ascholerChemeketa:
Updated to filter import * related warnings from pyflakes to avoid scary warnings in existing books.

PR #1397, opened 2023-05-10
Micro Parsons: Render blocks raw when all blocks are html in non-html problems

zihan21206:
RunestoneInteractive/rs#20 

Render blocks raw when micro Parsons find all blocks contain HTML, and when the language is not html.

Reason for asking all blocks to contain HTML:
a complete HTML can be used as a string in other languages, e.g. in Python, `re.match('<b>', '<b>content</b>')`

Two examples are shown below (included in test).

image

image

micro-Parsons dependency is updated in this PR, so `npm install` is needed.

### DIFF
--- a/runestone/hparsons/js/hparsons.js
+++ b/runestone/hparsons/js/hparsons.js
@@ -26,6 +26,7 @@ export default class HParsons extends RunestoneBase {
         this.randomize = $(orig).data("randomize") ? true : false;
         this.isBlockGrading = $(orig).data("blockanswer") ? true : false;
         this.language = $(orig).data("language");
+        this.renderRaw = false;
         if (this.isBlockGrading) {
             this.blockAnswer = $(orig).data("blockanswer").split(" ");
         }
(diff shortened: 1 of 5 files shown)


runestone/hparsons/js/hparsons.js
```diff
@@ -80,6 +81,22 @@ export default class HParsons extends RunestoneBase {
         this.originalBlocks = this.processSingleContent(code, '--blocks--').split('\n').slice(1,-1);
         this.hiddenSuffix = this.processSingleContent(code, '--hiddensuffix--');
         this.unittest = this.processSingleContent(code, '--unittest--');
+        // (for pretext) if all blocks can be parsed as html but language is not html,
+        //               ask micro parsons to render raw
+        if (this.language != 'html') {
```
bnmnetp, line 86: I don't understand the logic of checking for not html.  I think you should check for language == 'natural' or 'math'. Those are the cases where we want the html formatting to show through.  Other languages should be verbatim as you have it I think.  @rbeezer correct me if I'm wrong.

rbeezer:
Yes, an author's source should specify "natural" language, or a computer language.  (There might be a default for authors, perhaps for historical reasons.)

But in any event, the generated HTML for the problem should have an explicit attribute to that effect which the code here can react to unambiguously.

